Issue #19, opened 2022-11-03
implement crypto.GetRandomValues


oleiade:
Implement and test the crypto.getRandomValues() function. As described in [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues), and specified in the [specification](https://www.w3.org/TR/WebCryptoAPI/#Crypto-method-getRandomValues)

PR #21, opened 2022-11-03
Implement support for the crypto.getRandomValues operation

oleiade:
This PR addresses #19 

It adds an implementation for the `crypto.getRandomValues` operation to the extension.

The underlying implementation depends on the Go `rand/crypto.Read` method as a PRNG. This standard functionality relies on the underlying OS' pseudo-random number generator, such as `/dev/urandom` on Linux. 

This implementation suffers a few limitations that shouldn't be a deal-breaker, but for which I couldn't find any acceptable workaround at this day:
- The [specification indicates](https://github.com/web-platform-tests/wpt/blob/master/WebCryptoAPI/getRandomValues.any.js#L1) that only certain views (`Int8Array`, `Uint8Array`, `Int16Array`, `Uint16Array`, `Int32Array`, `Uint32Array` or `UInt8ClampedArray`) on the underlying `ArrayBuffer` should be accepted. I couldn't figure out how to do that in Goja. I'd be really grateful if you have any idea on that front 🙇🏻 
- The specification indicates that certain specific errors such as `QuotaExceededError` should be returned under certain conditions. At the moment I have defined specific errors the Go way, and I throw them using `common.Throw`; but that doesn't allow me to, for instance, `switch` on them in the context of a `catch` statement, as in:
```javascript
switch (err) {
  case (err isinstanceof QuotaExceededError) {
```

Let me know what you think 🙇🏻


webcrypto/crypto.go
```diff
@@ -1,16 +1,94 @@
...
+		common.Throw(c.vu.Runtime(), TypeMismatchError)
+	}
+
+	// 2.
+	// Obtain the length of the typed array, and throw a QuotaExceededError if
+	// it's too big, as specified in the [spec's] 10.2.1.2 paragraph.
+	// [spec]: https://www.w3.org/TR/WebCryptoAPI/#Crypto-method-getRandomValues
+	obj := typedArray.ToObject(c.vu.Runtime())
+	objLength := obj.Get("length").ToNumber().Export().(int64)
+	if objLength > 65536 {
```
olegbespalov, line 62: Could you please move the "magic" number `65536` to a const with some explicit name?

webcrypto/crypto.go
```diff
@@ -1,16 +1,94 @@
...
+// its cryptographic sense, not in its statistical sense).
+//
+// To guarantee enough performance, this implementation is not using a truly
+// random number generator, but is using a pseudo-random number generator
+// seeded with a value with enough entropy. We are using the golang
+// crypto/rand package, which uses the operating system's random number
+// generator.
+func (c *Crypto) GetRandomValues(typedArray goja.Value) goja.Value {
+	// 1.
+	// If the provided typedArray is not of an integer type (Int8Array, Uint8Array,
```
olegbespalov, line 43: I appreciate your comments (here and in many other places), but do you think we need them when the code explains itself? 

I mean, the lines below bellow are easy to understand and don't require any additional commenting. Comments have their cost, e.g., maintenance; they quickly become obsolete if somebody forgets to update them during some one-line fix. 

Generally, we have less text (code + comments) in the method, which is more accessible from the cognitive side to get the point.

webcrypto/crypto.go
```diff
@@ -1,16 +1,94 @@
...
+	// on Windows. This is the recommended way to generate random
+	// by the specification.
+	randomValues := make([]byte, objLength)
+	_, err := rand.Read(randomValues)
+	if err != nil {
+		common.Throw(c.vu.Runtime(), err)
+	}
+
+	for i := int64(0); i < objLength; i++ {
+		err := obj.Set(fmt.Sprintf("%d", i), randomValues[i])
```
olegbespalov, line 83: Nit: Maybe it's a pre-mature optimization, but please consider using `strconv.FormatInt` (https://pkg.go.dev/strconv#FormatInt), which is faster & has a better memory footprint.

webcrypto/crypto.go
oleiade: Fair point, wilco 👍🏻

webcrypto/crypto.go
oleiade: You're right, actually I meant to replace the comment with a `var ()` statement with all the checks to make the `if` statement clearer, will do just that 👍🏻

webcrypto/crypto.go
oleiade: Interesting! Today, I learned something. I always tend to reach out for `sprintf` out of habit from my years with C, but now that [I read about it](https://medium.com/swlh/bad-go-frivolous-sprintf-2ad28fedf1a0), it does make sense! Thanks for bringing that up 🙇🏻

olegbespalov (approved):

imiric (approved):
LGTM :clap: Just left some nitpicks.

As for the limitations you mentioned, I see you've resolved the first one. :thumbsup:

I think the second one is possible, but you'd need to first make the error object constructor available to the goja VM, and then use it in Go to create an instance. I took a glance at goja tests, but my goja-fu is not great, so I summon the wizardry of @mstoykov.

But technically, these are [`DOMException`](https://developer.mozilla.org/en-US/docs/Web/API/DOMException) errors, and exposing that wouldn't make sense for k6. We could expose them as separate constructors I guess, and choose to not follow the spec slightly. I think that would be better than the currently thrown plain object. Let's maybe create an issue for it and fix it later?

mstoykov:
> The specification indicates that certain specific errors such as QuotaExceededError should be returned under certain conditions. At the moment I have defined specific errors the Go way, and I throw them using common.Throw; but that doesn't allow me to, for instance, switch on them in the context of a catch statement, as in:

AFAIK this will require that you actually define this error and then make instances of them.

webcrypto/crypto_test.go
```diff
@@ -0,0 +1,175 @@
...
+	ts := newTestSetup(t)
+
+	gotScriptErr := ts.ev.Start(func() error {
+		_, err := ts.rt.RunString(`
+		var input = new Uint8Array(10);
+		var output = crypto.getRandomValues(input);
+
+		if (input.length != 10) {
+			throw new Error("input.length != 10");
+		}
```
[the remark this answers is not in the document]
oleiade: Hehe nice catch, it's a leftover 🙇🏻

webcrypto/crypto.go
```diff
@@ -26,10 +34,80 @@ type Crypto struct {
...
+	// 1.
+	if !isInt8Array &&
+		!isUint8Array &&
+		!isInt16Array &&
+		!isUint16Array &&
+		!isInt32Array &&
+		!isUint32Array &&
+		!isUint8ClampedArray {
+		common.Throw(c.vu.Runtime(), NewError(0, TypeMismatchError, "typedArray parameter isn't a TypedArray instance"))
+	}
```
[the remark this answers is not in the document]
oleiade: Thanks a lot for this, it sparked new ideas on how to approach it indeed 🙇🏻 What I ended up doing was:
- rewriting the `IsInstanceOf` method to[ take the target types as variadic args](https://github.com/grafana/xk6-webcrypto/pull/21/files#diff-ec290271c57c152690bdb3ccbaf1bd30f6d5cfec670c878752333a7d691e6bceR7), and iterate over them using your method instead.
- replacing this specific call you pointed out to[ use the new prototype of `IsInstanceOf`](https://github.com/grafana/xk6-webcrypto/pull/21/files#diff-80708ffc0f79a50c11b94ad8ecf62cf1bdab64d56848b224d1e2c630323e4aceR49)

Regarding the structure of the function itself, I'd rather keep it as is. At least during the first implementation run. Considering the small scope of the function, I find it more convenient to track the implementation's correctness to have inlined and mirroring the specs. I find that splitting in smaller functions will prove useful in later functionalities (encrypt, decrypt, etc), but for the functions in the `crypto` namespace, I would judge it overkill abstraction for now 👍🏻

Issue #20, opened 2022-11-03
implement randomUUID


oleiade:
Implement and test the `crypto.randomUUID()` function. As described in [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID).

PR #22, opened 2022-11-03
Implement support for the crypto.randomUUID operation

oleiade:
This PR addresses #20.

It adds an implementation for the `crypto.randomUUID` operation to the extension.

```javascript

export default function () {
  const myUUID = crypto.randomUUID();

  console.log(myUUID);
}
```

It implements the function according to the [Web Crypto API's specification](https://w3c.github.io/webcrypto/#Crypto-method-randomUUID). Please note that this feature is part of a restricted feature set of the API.

Under the hood, it uses [Google's UUID](https://pkg.go.dev/github.com/google/uuid#section-readme) package, version `v1.2.2` like in k6, to produce UUID values.

Let me know what you think 🙇🏻


olegbespalov (approved):

imiric (approved):
LGTM :+1:

I would DRY the tests a bit since there's a lot of repetition, but it's not a big deal.

oleiade:
@imiric thanks 🙇🏻 

I believe I mentioned it in the tests, but those are almost verbatim copies of the [Web Platform Tests](https://github.com/web-platform-tests/wpt/blob/master/WebCryptoAPI/randomUUID.https.any.js) covering the Web Crypto API. Thus the not dryness is not my fact 🥥

Issue #8, opened 2022-11-02
Implement subtle.digest()


oleiade:
Implement and test the SubtleCrypto.digest() function. As described in [MDN](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest), and specified in the [specification](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-digest)

PR #23, opened 2022-11-11
Implement support for the `crypto.subtle.digest` operation

oleiade:
This PR addresses #8 .

It adds an implementation for the `crypto.subtle.digest` operation to the extension.

```javascript

export default function () {
  const d = crypto.subtle.digest("SHA-256", "Hello, world!").then((hash) => {
    const h = buf2hex(hash);
    console.log(h);
  });
}

function buf2hex(buffer) {
  return [...new Uint8Array(buffer)]
    .map((x) => x.toString(16).padStart(2, "0"))
    .join("");
}
```

It implements the function according to the [Web Crypto API's specification][(https://w3c.github.io/webcrypto/#Crypto-method-randomUUID](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-digest)). Please note that this feature is part of a restricted feature set of the API.

Note that the [test suite](https://github.com/grafana/xk6-webcrypto/commit/8ecb0b3258b4e4610ce5e9cf47202f70b9228035#diff-d7be9ed254eb99f263669260bf5acaa640f7c2d12b66c16252fb3340bbddc5faR22) is inherited and adapted from the [WPT test suite](https://github.com/web-platform-tests/wpt/blob/master/WebCryptoAPI/digest/digest.https.any.js) dedicated to the `digest` operation.

## Checklist

- [ ] Although not absolutely mandatory, as we have the WPT test suite testing this end2end, I'd like to add more unit tests for this.

Let me know what you think 🙇🏻

### DIFF
--- a/examples/digest.js
+++ b/examples/digest.js
@@ -1,17 +1,22 @@
-import { sleep } from "k6";
 import { crypto } from "k6/x/webcrypto";
 
-console.log(JSON.stringify(crypto));
-
 export default function () {
-  const d = crypto.subtle.digest("SHA-256", "Hello, world!").then((hash) => {
-    const h = buf2hex(hash);
-    console.log(h);
-  });
+  crypto.subtle.digest("SHA-256", stringToArrayBuffer("Hello, world!")).then(
+    (hash) => {
+      console.log(arrayBufferToHex(hash));
+    },
+    (err) => {
+      throw err;
+    }
+  );
 }
 
-function buf2hex(buffer) {
+function arrayBufferToHex(buffer) {
   return [...new Uint8Array(buffer)]
     .map((x) => x.toString(16).padStart(2, "0"))
     .join("");
+}
+
+function stringToArrayBuffer(s) {
+  return Uint8Array.from(new String(s), (x) => x.charCodeAt(0));
 }
--- a/webcrypto/algorithm.go
+++ b/webcrypto/algorithm.go
@@ -4,7 +4,10 @@ import (
 	"crypto"
 	"fmt"
 	"hash"
+	"reflect"
 	"strings"
+
+	"github.com/dop251/goja"
 )
 
 // Algorithm represents
@@ -12,39 +15,10 @@ type Algorithm struct {
 	Name AlgorithmIdentifier `json:"name"`
 }
 
-// Ensure AesKeyGenParams implements the From interface.
-var _ From[map[string]interface{}, Algorithm] = Algorithm{}
-
-// From implements the From interface for Algorithm, and initializes the
-// Algorithm instance from a map[string]interface{}.
-func (a Algorithm) From(dict map[string]interface{}) (Algorithm, error) {
-	algorithm := Algorithm{}
-	nameFound := false
-
-	for key, value := range dict {
-		if strings.EqualFold(key, "name") {
-			name, ok := value.(string)
-			if !ok {
-				return Algorithm{}, NewError(0, NotSupportedError, "algorithm name is not a string")
-			}
-
-			name = strings.ToUpper(name)
-
-			if !IsAlgorithm(name) && !IsHashAlgorithm(name) {
-				return Algorithm{}, NewError(0, NotSupportedError, "algorithm name is not supported")
-			}
-
-			algorithm.Name = name
-			nameFound = true
-			break
-		}
-	}
-
-	if !nameFound {
-		return Algorithm{}, NewError(0, NotSupportedError, "algorithm name is not found")
-	}
-
-	return algorithm, nil
+// NormalizedName returns the normalized algorithm identifier.
+// It implements the NormalizedIdentifier interface.
+func (a Algorithm) NormalizedName() AlgorithmIdentifier {
+	return a.Name
 }
 
 // AlgorithmIdentifier represents the name of an algorithm.
@@ -160,45 +134,47 @@ func Hasher(algorithm HashAlgorithmIdentifier) (func() hash.Hash, error) {
 	return nil, NewError(0, ImplementationError, fmt.Sprintf("unsupported hash algorithm: %s", algorithm))
 }
 
+// NormalizedAlgorithm represents a normalized algorithm.
+type NormalizedAlgorithm interface {
+	NormalizedName() AlgorithmIdentifier
+}
+
 // normalize algorithm
 // normalizeAlgorithm(algorithm: string | Algorithm, op: string):
 
 // NormalizeAlgorithm normalizes the given algorithm following the algorithm described in the WebCrypto [specification].
 //
 // [specification]: https://www.w3.org/TR/WebCryptoAPI/#algorithm-normalization-normalize-an-algorithm
-func NormalizeAlgorithm(algorithm interface{}, op OperationIdentifier) (interface{}, error) {
-	// var initialAlg AlgorithmIdentifier
-	var initialAlg map[string]interface{}
-
-	switch alg := algorithm.(type) {
-	case string:
-		return NormalizeAlgorithm(map[string]interface{}{"name": alg}, op)
-	case map[string]interface{}:
-		// FIXME: this should call the NewAlgorithmFrom method instead
-		if _, ok := alg["name"]; !ok {
-			return Algorithm{}, NewError(0, SyntaxError, "algorithm name is not found")
-		}
-
-		if _, ok := alg["name"].(string); !ok {
-			return Algorithm{}, NewError(0, SyntaxError, "algorithm name is not a string")
-		}
-
-		initialAlg = alg
-	default:
-		return Algorithm{}, NewError(0, ImplementationError, "unsupported algorithm type")
-	}
-
+func NormalizeAlgorithm(rt *goja.Runtime, algorithm goja.Value, op OperationIdentifier) (NormalizedAlgorithm, error) {
 	// 1.
 	registeredAlgorithms, ok := supportedAlgorithms[op]
 	if !ok {
 		return Algorithm{}, NewError(0, ImplementationError, fmt.Sprintf("unsupported operation: %s", op))
 	}
 
-	// 2. 3. 4.
-	algName, ok := initialAlg["name"].(string)
-	if !ok {
-		return Algorithm{}, NewError(0, SyntaxError, "algorithm name is not a string")
+	// 2.
+	var initialAlg Algorithm
+	switch algorithm.ExportType().Kind() {
+	case reflect.String:
+		obj := rt.NewObject()
+		err := obj.Set("name", algorithm.Export())
+		if err != nil {
+			// 3.
+			return Algorithm{}, NewError(0, ImplementationError, "unable to convert the string argument to an object")
+		}
+		return NormalizeAlgorithm(rt, obj, op)
+	case reflect.Map, reflect.Struct:
+		err := rt.ExportTo(algorithm, &initialAlg)
+		if err != nil {
+			// 3.
+			return Algorithm{}, NewError(0, SyntaxError, "algorithm object is not a valid algorithm")
+		}
+	default:
+		return Algorithm{}, NewError(0, SyntaxError, "unsupported algorithm type")
 	}
+
+	// 4.
+	algName := initialAlg.Name
 
 	// 5.
 	var desiredType string
@@ -222,7 +198,6 @@ func NormalizeAlgorithm(algorithm interface{}, op OperationIdentifier) (interfac
 	}
 
 	// 6.
-	// FIXME: handle this case in later versions
 	err := NewError(0, ImplementationError, fmt.Sprintf("unsupported algorithm type: %s", desiredType))
 	return Algorithm{}, err
 }
--- a/webcrypto/subtle_crypto.go
+++ b/webcrypto/subtle_crypto.go
@@ -1,6 +1,10 @@
 package webcrypto
 
 import (
+	"crypto"
+	"errors"
+	"hash"
+
 	"github.com/dop251/goja"
 	"go.k6.io/k6/js/modules"
 )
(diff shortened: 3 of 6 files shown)


webcrypto/subtle_crypto.go
```diff
@@ -137,8 +141,87 @@ func (sc *SubtleCrypto) Verify(
...
+	// Validate that the value we received is either an ArrayBuffer, TypedArray, or DataView
+	// This uses the technique described in https://github.com/dop251/goja/issues/379#issuecomment-1164441879
+	var (
+		isArrayBuffer = IsInstanceOf(sc.vu.Runtime(), data, ArrayBufferConstructor)
+		isDataView    = IsInstanceOf(sc.vu.Runtime(), data, DataViewConstructor)
+		isTypedArray  = IsTypedArray(sc.vu.Runtime(), data)
+	)
+
+	if !isArrayBuffer && !isDataView && !isTypedArray {
+		reject(fmt.Errorf("data must be an ArrayBuffer, TypedArray, or DataView"))
```
olegbespalov, line 156: Here (and in other places below), we don't need `fmt.Errorf`. It's just a `errors.New`

webcrypto/subtle_crypto.go
```diff
@@ -137,8 +141,87 @@ func (sc *SubtleCrypto) Verify(
...
+
+	// The specification explicitly requires us to copy the underlying
+	// bytes held by the array buffer
+	bytesCopy := make([]byte, len(bytes))
+	copy(bytesCopy, bytes)
+
+	// 3.
+	normalizedAlgorithm, err := NormalizeAlgorithm(algorithm, OperationIdentifierDigest)
+	if err != nil {
+		// "if an error occurred, return a Promise rejected with NormalizedAlgorithm"
```
olegbespalov, line 186: :thinking: I spotted this in WebCrypto specification, but to be honest, I'm not sure if I get this comment. Maybe I'm (and imagine like any other reader of this code) missing some context that we can put here.

What does mean `rejected with NormalizedAlgorithm"`?

webcrypto/subtle_crypto.go
```diff
@@ -137,8 +141,87 @@ func (sc *SubtleCrypto) Verify(
...
+		reject(err)
+		return promise
+	}
+
+	alg, ok := normalizedAlgorithm.(Algorithm)
```
olegbespalov, line 191: Out of curiosity, why does `NormalizeAlgorithm` returns an interface, not the `Algorithm`?

webcrypto/subtle_crypto.go
oleiade: This comment is legacy and reminded me to change the older implementation. I'll drop it.

The Web Crypto specification is very cryptic in certain regards, indeed. It took me a long time and a lot of research to determine what that meant. Most of what I've done on that front is inspired by the [Deno implementation](https://github.com/denoland/deno/blob/main/ext/crypto/00_crypto.js). 

The algorithm normalization is essentially the step of converting an input `Object` or `string`, in a predefined format, into a (set of) internal structures that can be worked on while also ensuring to fail if any validation steps fail. It's a step that's quite easy in a dynamic language but was quite hard to do in a static language. I'm still not happy with the results, but it will do for now :)

I'll add a comment here to explain just that 👍🏻

webcrypto/subtle_crypto.go
oleiade: Good catch 👍🏻

webcrypto/subtle_crypto.go
oleiade: I believe the reason for that will become clearer in the further PRs you look at. In the context of digest not so much 👍🏻 

In short there are many kinds of `Algorithm`. It could be an instance of `KeyAlgorithm`, an `RSAHashedKeyAlgorithm`, or one of a couple of others. And each actions and their own set of data structures will come with different expectations. For instance RSA expects the Algorithm field to hold a `RSAHashedKeyAlgorithm` flavor, which exposes the hash algorithm used.

 I considered using Generics to handle that case, but in my experience at this point with the lib and with dealing with goja, it's easier and more convenient to do runtime checking and casting rather than trying to be too clever 😞

olegbespalov: That I can understand, but still, it sounds like instead of the null interface, there can be some specific interface that can combine the most generic representation of the algorithm.

As an example bellow could look like this:
```golang
type NormalizedAlgorithm interface {
   Algorithm() string 
}


func NormalizeAlgorithm(algorithm interface{}, op OperationIdentifier) (NormalizedAlgorithm, error) {
 ...
}

switch alg.Algorithm() {}

```
```
Otherwise, whenever you use this method, you should do casting.

Comment (oleiade) [reply webcrypto/subtle_crypto.go:191]: I believe you're onto something 😄 I need to think about it a bit more, and I'll get back to this comment.

In the meantime, here's how `NormalizeAlgorithm` is used in a more "complex" and "standard" use case. Digest is a limited and quite specific subset indeed.

**Review by imiric — CHANGES_REQUESTED** [review]
Hey, great work on this, but I think you're still working on it, right? Since the test doesn't actually work for me, even after converting the data to a TypedArray, and the CI fails. It would be better if this PR was converted to draft if so.

Still, I left some comments from the first look.

File: webcrypto/subtle_crypto_test.go [hunk 4]
```diff
@@ -0,0 +1,178 @@
...
+//
+//nolint:dupword
+func TestSubtleDigest(t *testing.T) {
+	t.Parallel()
+
+	ts := newTestSetup(t)
+
+	gotScriptErr := ts.ev.Start(func() error {
+		_, err := ts.rt.RunString(`
+			var subtle = crypto.subtle;
```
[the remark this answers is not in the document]
oleiade: Very good point, I had actually considered doing it, but by bringing it up, you nudged me to do it 🙇🏻 👍🏻

examples/digest.js
```diff
@@ -1,8 +1,5 @@
-import { sleep } from "k6";
 import { crypto } from "k6/x/webcrypto";
 
-console.log(JSON.stringify(crypto));
-
 export default function () {
   const d = crypto.subtle.digest("SHA-256", "Hello, world!").then((hash) => {
```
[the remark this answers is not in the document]
oleiade: Indeed, this is a limitation at the moment. The specification relies on the fact that users will feed their ArrayBuffers themselves, using other specified constructs such as `TextEncoder`. However, I'd like to leave encoding of strings out of scope in this first iteration for mainly two reasons:
- Implementing the specification involves enough complexity, and I'd prefer not to add immediate deviation from it in this specific PR.
- It is possible to build small functions in pure Javascript doing the string encoding into array buffers without resorting to `TextEncoder`: https://github.com/grafana/xk6-webcrypto/pull/24/files#diff-4f6907dc6ec2ae10dc3b26dae0ff0135703a771ffb69fc756505b7704f836d76R40

I would open an issue to keep that topic on the roadmap, though, and I would like to address it in some way indeed separately. Let me know if that makes sense for you 🙇🏻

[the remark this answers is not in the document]
oleiade: I've updated the example to showcase it 👍🏻

webcrypto/subtle_crypto.go
```diff
@@ -137,8 +141,87 @@ func (sc *SubtleCrypto) Verify(
...
+		switch alg.Name {
+		case Sha1:
+			hash = crypto.SHA1.New()
+		case Sha256:
+			hash = crypto.SHA256.New()
+		case Sha384:
+			hash = crypto.SHA384.New()
+		case Sha512:
+			hash = crypto.SHA512.New()
+		default:
```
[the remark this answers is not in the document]
oleiade: In principle, I agree. However, the traits of the types that could be returned by `NormalizeAlgorithm` will vary quite a bit. In certain scenarios, it will need to return something that can be used to generate a key, in other scenarios, something that can be used to export a key or encrypt. Hence, returning an interface might turn out complex, if not impossible. I'm experimenting with it, and I'll get back to you folks 👍🏻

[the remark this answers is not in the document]
oleiade: I've substantially cleaned the algorithm normalization up. Plus, I added a `NormalizedAlgorithm` interface which allows doing exactly what you requested. I'm not sure it will survive the "fire baptism" in larger context (`generateKey`, `encrypt`, etc), but the design is already better like this, and it will force us to find better designs down the road indeed 🙇🏻 👍🏻

webcrypto/subtle_crypto.go
oleiade: I've substantially cleaned the algorithm normalization up. Plus, I added a `NormalizedAlgorithm` interface which allows doing exactly what you requested. I'm not sure it will survive the "fire baptism" in a larger context (`generateKey`, `encrypt`, etc), but the design is already better like this, and it will force us to find better designs down the road indeed 🙇🏻 👍🏻

webcrypto/subtle_crypto_test.go
```diff
@@ -0,0 +1,178 @@
...
+			for (var i=0; i<1024; i++) {
+				sourceData.long.set(sourceData.medium, i * sourceData.medium.byteLength);
+			}
+
+			var digestedData = {
+				"sha-1": {
+					empty: new Uint8Array([218, 57, 163, 238, 94, 107, 75, 13, 50, 85, 191, 239, 149, 96, 24, 144, 175, 216, 7, 9]),
+					short: new Uint8Array([201, 19, 24, 205, 242, 57, 106, 1, 94, 63, 78, 106, 134, 160, 186, 101, 184, 99, 89, 68]),
+					medium: new Uint8Array([229, 65, 6, 8, 112, 235, 22, 191, 51, 182, 142, 81, 245, 19, 82, 104, 147, 152, 103, 41]),
+					long: new Uint8Array([48, 152, 181, 0, 55, 236, 208, 46, 189, 101, 118, 83, 178, 191, 160, 30, 238, 39, 162, 234])
```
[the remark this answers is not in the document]
oleiade: I see where you come from. I had the same initial reaction. But I'd much rather keep the tests as-is. At least for now. The main reasons on my side are:
- Those are the official test data sets. I hate the authority argument, but regarding cryptography, I accept it.
- In this current iteration, we accept only ArrayBuffers and views on ArrayBuffers as input. While I think we need to think of a strategy forward to encode/decode strings in this module, I think it should be addressed as a separate topic.
- Mostafa pointed out to me in the discussion about cryptography I had with him that this is the standard way to test implementations of cryptographic algorithms. Some specifications, like NIST's AES cipher, for instance, publish vectors of input bytes and their expected output.

If you have a strong feeling towards having string-based tests, please let me know, and we could still add dedicated tests on top of those 🙇🏻

examples/digest.js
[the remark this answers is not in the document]
imiric: Thanks, but wouldn't adding support for string be a trivial change? Forcing users to use a helper encoding for such a common scenario doesn't seem friendly. That's assuming they can come up with it on their own, or find it in our examples or documentation.

I wouldn't call it a deviation from the spec, just augmenting it to support a very common usage scenario.

So I would prefer if we add it, though it can be done in a separate PR.

webcrypto/subtle_crypto.go
```diff
@@ -137,8 +141,87 @@ func (sc *SubtleCrypto) Verify(
...
+	// to a Goja ArrayBuffer, and then getting its underlying Go slice
+	// by calling the `Bytes()` method.
+	//
+	// Doing so conviniently also copies the underlying buffer, which
+	// is required by the specification.
+	// See https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-digest
+	asObject := data.ToObject(sc.vu.Runtime())
+	arrayBuffer, ok := asObject.Export().(goja.ArrayBuffer)
+	if !ok {
+		reject(fmt.Errorf("could not get ArrayBuffer from data"))
```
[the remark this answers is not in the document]
imiric: I'm familiar with TypedArrays, but I was hoping we could get their ArrayBuffer using the Goja Go API, instead of doing it in JS via the `buffer` property. It would be safer if we didn't rely on duck typing :sweat_smile: If that's not possible, then I suppose it's the only thing we can do.

Though your latest change breaks passing an actual `ArrayBuffer` instance, which should be supported according to the spec, so we should type assert on both `data` and `data.ToObject(rt).Get("buffer")`.

Also, let's check for `null`/`undefined` as well.

webcrypto/subtle_crypto_test.go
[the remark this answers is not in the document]
imiric: I'm not suggesting to avoid testing with bytes as input, but to have tests that use readable strings as a starting point, decode them to byte arrays, and from that create the TypedArray that is passed to `digest()`. Essentially, using the `stringToArrayBuffer()` helper you added in the example. As I can't do the encoding in my head :sweat_smile:, seeing a bunch of byte arrays is useless, precisely from a maintenance perspective. So I was hoping we could use ASCII-friendly strings instead.

But you're right that there's a lot of value in reusing the official WPT tests, so OK, let's leave this as is.

webcrypto/subtle_crypto.go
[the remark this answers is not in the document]
oleiade: I see what you mean. I understand that doing it in Go is not possible now. If I remember correctly, when I experimented with it, treating the value as a `goja.ArrayBuffer` led to a behavior where the ArrayBuffer was empty.  The solution I adopted here is based on this discussion: https://github.com/dop251/goja/issues/379.

Good catch on the `ArrayBuffer` breaking change 👍🏻 I'll update it accordingly 🙇🏻

examples/digest.js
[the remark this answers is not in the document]
oleiade: I'm definitely not opposed to that, but I would indeed do it in a separate PR to avoid adding scope and complexity to this one 👍🏻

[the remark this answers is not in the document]
oleiade: tracked in #29

examples/digest.js
```diff
@@ -1,17 +1,22 @@
...
-    console.log(h);
-  });
+  crypto.subtle.digest("SHA-256", stringToArrayBuffer("Hello, world!")).then(
+    (hash) => {
+      console.log(arrayBufferToHex(hash));
+    },
+    (err) => {
+      throw err;
+    }
+  );
```
mstoykov, line 11: nit: probably should be using the `async`/`await` syntax at this point in time.

webcrypto/algorithm.go
```diff
@@ -160,46 +134,48 @@ func Hasher(algorithm HashAlgorithmIdentifier) (func() hash.Hash, error) {
...
 		return Algorithm{}, NewError(0, ImplementationError, fmt.Sprintf("unsupported operation: %s", op))
 	}
 
-	// 2. 3. 4.
-	algName, ok := initialAlg["name"].(string)
-	if !ok {
-		return Algorithm{}, NewError(0, SyntaxError, "algorithm name is not a string")
+	// 2.
+	var initialAlg Algorithm
+	switch algorithm.ExportType().Kind() {
```
mstoykov, line 157: I am not really a fan of using `ExportType` as IME it is really ... fragile.

Usually I would recommend actually checking the prototype of the value, but this is fine for now.

webcrypto/algorithm.go
```diff
@@ -222,7 +198,6 @@ func NormalizeAlgorithm(algorithm interface{}, op OperationIdentifier) (interfac
 	}
 
 	// 6.
-	// FIXME: handle this case in later versions
 	err := NewError(0, ImplementationError, fmt.Sprintf("unsupported algorithm type: %s", desiredType))
```
mstoykov, line 201: Why not keep the comment?

webcrypto/subtle_crypto.go
mstoykov: I feel like this comment *made* it worse :rofl: 

The specification says:
```
Let normalizedAlgorithm be the result of [normalizing an algorithm](https://www.w3.org/TR/WebCryptoAPI/#dfn-normalize-an-algorithm), with alg set to algorithm and op set to "digest".

If an error occurred, return a Promise rejected with normalizedAlgorithm.
```

And as the Specification does not have multiple returns `normalizedAlgorithm` is just the result of the operation which *can* be an error (which "occurred"). 

So from that I would recommend *not* having a comment as that IMO makes it worse 
```suggestion
```

oleiade:
Thanks a lot for taking the time to review it @mstoykov much appreciated. I'll merge the PR for now, to move forward, and will address your comments separately 🙇🏻

webcrypto/algorithm.go
oleiade: This was a rebase mistake, for sure 🤦🏻

Issue #13, opened 2022-11-02
Implement subtle.exportKey()


oleiade:
Implement and test the SubtleCrypto.exportKey() function. As described in [MDN](), and specified in the [specification]()

PR #26, opened 2022-11-29
Implement support for the `subtle.generateKey` operation

oleiade:
This PR addresses #13 and implements the `subtle.generateKey` operation.

It implements the function according to the [W3C specification](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-generateKey), and offers [the following API](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/generateKey).

The PR contains extensive usage examples offered by the functionalities of this PR, but here's a quick, illustrative example, of an ECDSA key pair generation:
```javascript

export default function () {
  crypto.subtle
    .generateKey({ name: "ECDSA", namedCurve: "P-256" }, true, [
      "sign",
      "verify",
    ])
    .then((key) => {
      console.log(JSON.stringify(key));
    });
}
```

I'm sure there are many potential improvements possible left for this, and I'm looking forward to hearing from you 🙇🏻 

@mostafa, I added you as an optional reviewer; if you have time, could you please take a quick look to ensure I haven't done something wrong on the key generation part? The code for those live in the `aes.go`, `rsa.go`, `elliptic_curve.go` and `hmac.go` files (`GenerateKey` functions).

### DIFF
--- a/webcrypto/aes.go
+++ b/webcrypto/aes.go
@@ -16,7 +16,7 @@ type AesKeyAlgorithm struct {
 	Length int `json:"length"`
 }
 
-// AESKeyGenParams represents the object that should be passed as
+// AesKeyGenParams represents the object that should be passed as
 // the algorithm parameter into `SubtleCrypto.generateKey`, when generating
 // an AES key: that is, when the algorithm is identified as any
 // of AES-CBC, AES-CTR, AES-GCM, or AES-KW.
@@ -48,7 +48,7 @@ func (a AesKeyGenParams) From(dict map[string]interface{}) (AesKeyGenParams, err
 				return AesKeyGenParams{}, NewError(0, NotSupportedError, fmt.Sprintf("algorithm %s is not supported", name))
 			}
 
-			a.Name = AlgorithmIdentifier(name)
+			a.Name = name
 			nameFound = true
 			continue
 		}
@@ -80,14 +80,24 @@ func (a AesKeyGenParams) From(dict map[string]interface{}) (AesKeyGenParams, err
 var _ KeyGenerator = &AesKeyGenParams{}
 
 // GenerateKey generates a new AES key.
-func (a *AesKeyGenParams) GenerateKey(rt *goja.Runtime, extractable bool, keyUsages []CryptoKeyUsage) (goja.Value, error) {
-	if a.Algorithm.Name != AESCbc && a.Algorithm.Name != AESCtr && a.Algorithm.Name != AESGcm && a.Algorithm.Name != AESKw {
+func (a *AesKeyGenParams) GenerateKey(
+	rt *goja.Runtime,
+	extractable bool,
+	keyUsages []CryptoKeyUsage,
+) (goja.Value, error) {
+	var (
+		isCBC = a.Algorithm.Name == AESCbc
+		isCTR = a.Algorithm.Name == AESCtr
+		isGCM = a.Algorithm.Name == AESGcm
+		isKW  = a.Algorithm.Name == AESKw
+	)
+	if !isCBC && !isCTR && !isGCM && !isKW {
 		return nil, NewError(0, ImplementationError, "invalid algorithm")
 	}
 
 	// 1.
 	for _, usage := range keyUsages {
-		if strings.EqualFold(string(a.Algorithm.Name), AESKw) {
+		if strings.EqualFold(a.Algorithm.Name, AESKw) {
 			switch usage {
 			case WrapKeyCryptoKeyUsage, UnwrapKeyCryptoKeyUsage:
 			default:
@@ -110,8 +120,7 @@ func (a *AesKeyGenParams) GenerateKey(rt *goja.Runtime, extractable bool, keyUsa
 	// 3.
 	// FIXME: verify if there are different constraints on the key format depending on the AES flavor
 	randomKey := make([]byte, a.Length/8)
-	_, err := rand.Read(randomKey)
-	if err != nil {
+	if _, err := rand.Read(randomKey); err != nil {
 		// 4.
 		return nil, NewError(0, OperationError, "could not generate random key")
 	}
--- a/webcrypto/aes.go
+++ b/webcrypto/aes.go
@@ -1,0 +1,146 @@
+package webcrypto
+
+import (
+	"crypto/rand"
+	"fmt"
+	"strings"
+
+	"github.com/dop251/goja"
+)
+
+// AesKeyAlgorithm represents the AES algorithm.
+type AesKeyAlgorithm struct {
+	KeyAlgorithm
+
+	// The length, in bits, of the key.
+	Length int `json:"length"`
+}
+
+// AESKeyGenParams represents the object that should be passed as
+// the algorithm parameter into `SubtleCrypto.generateKey`, when generating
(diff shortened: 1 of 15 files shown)


webcrypto/aes.go
```diff
@@ -0,0 +1,155 @@
...
+	extractable bool,
+	keyUsages []CryptoKeyUsage,
+) (goja.Value, error) {
+	var (
+		isCBC = a.Algorithm.Name == AESCbc
+		isCTR = a.Algorithm.Name == AESCtr
+		isGCM = a.Algorithm.Name == AESGcm
+		isKW  = a.Algorithm.Name == AESKw
+	)
+	if !isCBC && !isCTR && !isGCM && !isKW {
```
olegbespalov, line 94: This validation should be in the constructor method, not the `GenerateKey`. Otherwise, we're breaking concerns :cry: 

So the constructor should produce a valid object (`AesKeyGenParams`), and later any operation on it should be executed safely without additional checks.

webcrypto/elliptic_curve.go
```diff
@@ -1,5 +1,213 @@
...
+	}
+
+	if !namedCurveFound {
+		return EcKeyGenParams{}, NewError(0, SyntaxError, "missing elliptic curve name")
+	}
+
+	return params, nil
+}
+
+// Ensure AesKeyGenParams implements the From interface.
```
olegbespalov, line 110: That's why most comments are evil :sweat_smile:

webcrypto/subtle_crypto.go
```diff
@@ -160,8 +160,47 @@ func (sc *SubtleCrypto) Digest(algorithm goja.Value, data interface{}) *goja.Pro
...
+		reject(err)
+		return promise
+	}
+
+	// 5.
+	go func() {
+		var value goja.Value
+		var err error
+
+		// FIXME: params could be an interface here, depending on whatever underlying
```
olegbespalov, line 179: To continue the discussion started in https://github.com/grafana/xk6-webcrypto/pull/23#discussion_r1034378372

To be honest, it's pretty clear right now that we need an additional interface, and it can have the name `Algorithm` since now this struct does almost nothing, so it can be an internal `algorithm` with only `name` property inside.

```golang
type Algorithm interface {
    Name() string
    GenerateKey(*goja.Runtime, []CryptoKeyUsage) (goja.Value, error)
}
```

And AES, HMAC, and so on should just implement this interface.

olegbespalov (changes requested):
I did an initial round of review, but I believe the current approach should be adjusted by introducing the Algorithm interface, a constructor for specific algorithms, and so on.

webcrypto/aes.go
oleiade: You're right 👍🏻

webcrypto/elliptic_curve.go
oleiade: Hehe, I'm gonna blame Copilot here 😄 🐱 👍🏻

webcrypto/subtle_crypto.go
oleiade: Yes, I think you're right, I'll give it a try, and rebase the other PRs (encrypt, decrypt, and incoming import/export) to see if the ideas expands nicely and if it introduces breaking changes down the road 👍🏻

webcrypto/aes.go
```diff
@@ -0,0 +1,155 @@
...
+
+	// Length holds (a Number) the length of the key, in bits.
+	Length int `json:"length"`
+}
+
+// Ensure AesKeyGenParams implements the From interface.
+var _ From[map[string]interface{}, AesKeyGenParams] = AesKeyGenParams{}
+
+// From fills the AesKeyGenParams from the given dictionary object.
+func (a AesKeyGenParams) From(dict map[string]interface{}) (AesKeyGenParams, error) {
```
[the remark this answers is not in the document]
oleiade: As we've discussed during our call, I removed the `From` interface and its implementations in favor of constructors. I've also added a `Normalizer` and `Validator` set of interfaces to handle the local validation and shaping of the data.

I'll resolve this conversation; feel free to un-resolve it if you have any further comments 👍🏻

webcrypto/elliptic_curve.go
oleiade: I addressed it here, and in the other relevant place. Looking back at it, I find that although the comment needed to be fixed indeed, it is still valuable as a lot of Go developers don't know this syntax (I've actually seen question about it quite often) 👍🏻

webcrypto/subtle_crypto.go
oleiade: A `KeyGenerator` interface now enforces `*KeyGenParams` structs to implement the method. The `GenerateKey` method casts the normalized algorithm to a `KeyGenerator`  and is much simplified.

webcrypto/aes.go
```diff
@@ -0,0 +1,159 @@
...
+
+// GenerateKey generates a new AES key.
+//
+// It implements the CryptoKeyGenerator interface.
+func (a AesKeyGenParams) GenerateKey(
+	// rt *goja.Runtime,
+	extractable bool,
+	keyUsages []CryptoKeyUsage,
+) (*CryptoKey[[]byte], error) {
+	// 1.
```
mostafa, line 101: What are these comments?

webcrypto/elliptic_curve.go
```diff
@@ -1,5 +1,214 @@
...
+	// in the Go standard library, so we protect our users from ourselves (and themselves),
+	// and refuse to generate ECDH keys until it is implement in the Go stdlib.
+	//
+	// As of November 2022, a proposal has been accepted, and ECDH implementation into
+	// the golang standard library [is in progress].
+	//
+	// [is in progress]: https://github.com/golang/go/issues/52221
+	if e.Algorithm.Name == ECDH {
+		return nil, NewError(0, NotSupportedError, "ECDH key generation is not supported, yet")
+	}
```
mostafa, line 129: Implemented in 1.20: https://pkg.go.dev/crypto/ecdh

mostafa (commented):
LGTM!

webcrypto/aes.go
oleiade: Those are the specification's steps 😸 Deno does that throughout their implementation of specs, and as a code reader, I found it very useful: "why does it do X, oh, wait I can just check the step in the specs" 🙇🏻

webcrypto/elliptic_curve.go
oleiade: Indeed, I found the proposal as I was implementing, but considering that we're still with 1.18, and 1.20 and is not out yet, I'll wait for it to be out, and for k6 to use it before I activate support for ECDH 👍🏻 :bow

webcrypto/aes.go
mostafa: It would be nice to mention this somewhere in the README.

webcrypto/elliptic_curve.go
mostafa: Fair enough.

webcrypto/aes.go
oleiade: Good point! Will make sure to add one!

PR #30, opened 2023-03-20
[1/3] Implement `generateKey` with support for AES algorithms

oleiade:
## What this is

This PR implements to support the [`subtle.crypto.generateKey`](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-generateKey) operation.  To keep Pull Requests as small and manageable as possible, **It only implements support for the AES family of algorithms**.

The operation allows the generation of a new random cryptographic key, which can later be used to perform encryption/decryption, or signature/verification operations (implemented in further PRs).

🟢 While the PR shows +1000 changes, 700 of them are from the official test suites, which **do not need to be reviewed** 🟢 

### Test suite

While the PR might look big at first glance, you can safely disregard and ignore the files in the `webcrypto/tests` folder. Those files are directly imported and slightly adapted from [the official test suite](https://github.com/web-platform-tests/wpt/tree/master/WebCryptoAPI) so that they are runnable in k6, but we haven't written those originally. We won't make any functional changes to them, as they are our safety-net, ensuring that our implementation matches the specifications expectation (and that what we do is 🥁 **safe**).

### Implementation

We have revamped quite a bit of the initial types, and data structure laid out during the first phases of the project. We decided to rely much more on `goja.Value` at the interface level than before and have refactored our types to limit their use of generics to where they were strictly useful/necessary/convenient.

## References

This PR is the first in a chain of 3 implementing support for key generation (1.), [import/export](https://github.com/grafana/xk6-webcrypto/pull/31#30) (2.), and [encryption/decryption](https://github.com/grafana/xk6-webcrypto/pull/32) using them. We intend to merge these in order in master: 1->2->3.


webcrypto/key.go
```diff
@@ -37,43 +30,24 @@ type CryptoKey[H KeyHandle] struct {
...
-
-// KeyGenerator is an interface that represents a cryptographic key generator.
-type KeyGenerator interface {
-	GenerateKey(rt *goja.Runtime, extractable bool, keyUsages []CryptoKeyUsage) (goja.Value, error)
-}
-
-// KeyHandle represents the underlying type of a key data handle.
-type KeyHandle interface {
-	[]byte | crypto.PrivateKey | crypto.PublicKey
+	handle interface{}
```
imiric, line 45:
```suggestion
	handle any
```

You're already using plenty of generics, so why not here? :blush:

(BTW, when will we have that generics workshop? :wink: :pray:)

webcrypto/aes.go
```diff
@@ -0,0 +1,84 @@
...
+func (akgp *AesKeyGenParams) GenerateKey(
+	extractable bool,
+	keyUsages []CryptoKeyUsage,
+) (*CryptoKey[AesKeyAlgorithm], error) {
+	for _, usage := range keyUsages {
+		switch usage {
+		case WrapKeyCryptoKeyUsage, UnwrapKeyCryptoKeyUsage:
+			continue
+		case EncryptCryptoKeyUsage, DecryptCryptoKeyUsage:
+			if akgp.Algorithm.Name != AESKw {
```
imiric, line 30: Could you add a comment here explaining why we're excluding AES-KW?

Should the error message be more informative in that case?

imiric (approved):
LGTM, great work all around! :clap: Just left minor suggestions.

A couple of general thoughts:

- I would've preferred to see the refactoring commits in a separate PR to make reviewing easier, but it's not a big deal.

- I suppose key generation is currently not that useful, even after they're exportable in #31, since they can't be written to local disk without an extension like xk6-file. We are planning to make this possible via the newly planned File API, right? Until then, users would have to transfer them over the network, which is not ideal.

webcrypto/tests/generateKey/failures.js
```diff
@@ -0,0 +1,229 @@
...
+// There are a lot of combinations of possible parameters for both
+// success and failure modes, resulting in a very large number of tests
+// performed.
+
+
+// Setup: define the correct behaviors that should be sought, and create
+// helper functions that generate all possible test parameters for
+// different situations.
+
+    // NOTE @oleiade: We quoted the value of the resultType property when its value is CryptoKey
```
mstoykov, line 32: How is the non quoted `CryptoKey` supposed to work, if you know?

webcrypto/tests/generateKey/failures.js
```diff
@@ -0,0 +1,229 @@
...
+        return result;
+    }
+
+    // Test that a given combination of parameters results in an error,
+    // AND that it is the correct kind of error.
+    //
+    // Expected error is either a number, tested against the error code,
+    // or a string, tested against the error name.
+    function testError(algorithm, extractable, usages, expectedError, testTag) {
+        return crypto.subtle.generateKey(algorithm, extractable, usages)
```
mstoykov, line 85: Here you removed `promise_test` invocation.

Wouldn't it better if this is not the case and we actually make `promise_test` work?

This might be done after some initial PR merging, but I do really prefer if we are checking ou the *actual* tests and running them without changes as much as possible.

If we need to change something, I guess it might be better to have the changes scripts with something like `ed`/`sed` or `.patch` files so those are better handled.

webcrypto/aes.go
mstoykov: Seems that it isn't supported by the stdlib https://github.com/golang/go/issues/27599

webcrypto/aes.go
```diff
@@ -0,0 +1,84 @@
...
+	key.Usages = keyUsages
+
+	// Set key handle to our random key.
+	key.handle = randomKey
+
+	// We apply the generateKey 8. step here, as we return a goja.Value
+	// instead of a CryptoKey(Pair).
+	if key.Usages == nil || len(key.Usages) == 0 {
+		return nil, NewError(0, SyntaxError, "the keyUsages argument must contain at least one valid usage for the algorithm")
+	}
```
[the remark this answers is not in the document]
mstoykov: I don't actually see anything in the specification about this ... so should we actually even error on this?

webcrypto/algorithm.go
```diff
@@ -322,3 +116,64 @@ const (
...
+
+	if v.ExportType().Kind() == reflect.String {
+		algorithmString, ok := v.Export().(string)
+		if !ok {
+			return Algorithm{}, NewError(0, ImplementationError, "algorithm cannot be interpreted as a string")
+		}
+
+		algorithm.Name = algorithmString
+
+		return algorithm, nil
```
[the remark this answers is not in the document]
mstoykov: The thing is that the specification (which I would prefer to follow to the letter whenever possible) actually explicitly states that this is how it should work.

Also in those places you linekd we again call `normalizeAlgorithm` and then it actually (should) *not* go in this if, so it will do the checks below.

webcrypto/subtle_crypto.go
```diff
@@ -234,8 +234,37 @@ func (sc *SubtleCrypto) Digest(algorithm goja.Value, data goja.Value) *goja.Prom
...
+		reject(err)
+		return promise
+	}
+
+	go func() {
+		switch normalized.Name {
+		case AESCbc, AESCtr, AESGcm, AESKw:
+			params := &AesKeyGenParams{
+				Algorithm: normalized,
+				Length:    algorithm.ToObject(sc.vu.Runtime()).Get("length").ToInteger(),
```
mstoykov, line 250: This is accessing the goja runtime not on the event loop. While this particualr case *might* not blow up in this particular case it is arguably better to just get the `length` before the goroutine or have a differnt way to acquire it that doesn't not access goja objects off the event loop

webcrypto/subtle_crypto_test.go
```diff
@@ -26,3 +27,47 @@ func TestSubtleDigest(t *testing.T) {
...
+			require.NoError(t, err)
+
+			_, err = ts.rt.RunString(`run_test()`)
+
+			return err
+		})
+
+		assert.NoError(t, gotScriptErr)
+	})
+}
```
mstoykov, line 73: Wouldn't this better with some kind of helper that just takes the files to run? Or runs all files within a directory like we(and goja) do for tc39 tests?

examples/generateKey/generateKey-aes.js
```diff
@@ -0,0 +1,17 @@
+import { crypto } from "k6/x/webcrypto";
+
+export default async function () {
+    const key = await crypto.subtle.generateKey(
+        {
+            name: "AES-CBC",
+            length: "256"
```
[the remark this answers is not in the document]
oleiade: Good catch 👀 🙇🏻

webcrypto/key.go
oleiade: I've actually switched back to using `any`, even for the `algorithm` attribute. As I've found out in later PRs, you cannot access the goja runtime from the promise callback without creating a race condition. Yet, most operations that were made easier and more robust by generics (we already know that in that case, we expect a params object describing a Aes Ctr algorithm so we can already use that specific type for our crypto keys) did rely on the ability to use the `Runtime().ExportTo` operation; which in turn forced me to do everything twice, just for the sake of generics.

We might find a workaround at a later point in time, and I'd be keen to bring generics back, then. However, at this point, they were really just forcing the code to be more complicated than necessary 😞

oleiade: Also, I'd be keen to make that now infamous workshop 🎉 but I guess I'd need to put up some specific, concrete, use case for it: WebCrypto was close 🤞🏻

webcrypto/aes.go
oleiade: I think I added one in another place, but will do it here too 👍🏻 

As Mihail pointed out, the reason is really that there is no official support for it from Go. In the previous implementation I wrote one myself, but didn't really trusted it.

webcrypto/tests/generateKey/failures.js
oleiade: I'm not 100% sure. However, considering the Web Platform Tests suite works under the assumption its run in a browser, they must also work under the assumption that `CryptoKey` is already exported in the top-level scope?

If I try in the Safari (webkit) web inspector to type `CryptoKey`, it seems to be natively there, without the need to import it 🤔 

```
> CryptoKey
function CryptoKey() {
    [native code]
} = $2
```

mstoykov: So at least in theory if we do `globalObject.CryptoKey = require("k6/experimental/webcrypto").Cryptokey` it should work? (I am pretty sure I made a typo somewhere in the code)

If so this might be one more thing that we don't really want to change in the test script but instead fix in the code runnign the test, by setting the environment as the test expect.

webcrypto/algorithm.go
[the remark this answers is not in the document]
imiric: Agreed that we should closely follow the spec, in which case this doesn't seem to be doing that.

Notice that [the spec](https://www.w3.org/TR/WebCryptoAPI/#algorithm-normalization-normalize-an-algorithm) says:

> If alg is an instance of a DOMString:
>   Return the result of running the [normalize an algorithm](https://www.w3.org/TR/WebCryptoAPI/#dfn-normalize-an-algorithm) algorithm, with the alg set to a new [Algorithm](https://www.w3.org/TR/WebCryptoAPI/#dfn-Algorithm) dictionary whose [name](https://www.w3.org/TR/WebCryptoAPI/#dfn-KeyAlgorithm-name) attribute is alg, and with the op set to op.

I.e. we should return the result of `normalizeAlgorithm()` here, passing an object instead. It might help if implement the algorithm in a clean function, free of any goja-isms.

My point is that we skip the validation if it's a string, which then makes us return `NotSupportedError` in other places where we don't need to. It's a simplification that the spec'd algorithm accounts for.

BTW, if we really want to be strict about following the spec, `normalizeAlgorithm()` should take an `op` (operation) argument, and return it in the normalized version, which we currently don't do. It seems this is a simplification for behavior we don't need, but in the case of validating a string, it's clearly missing from the implementation.

webcrypto/aes.go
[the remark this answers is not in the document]
oleiade: You're right folks, this was actually meant to live at the `generateKey` level, as the step [8 indeed](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-generateKey).  I've moved it [there](https://github.com/grafana/xk6-webcrypto/pull/30/files#diff-9e8a31e9776cf152256880a32e1c5751bf67322c23c0cb096d71cf0633d8fe14R263) 👍🏻

webcrypto/algorithm.go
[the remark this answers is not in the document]
oleiade: Hehe, I've spent literally hours staring at this specs in the last months, I guess I could almost recite them by heart at this point 🤝 😄 You're definitely right, I forgot to return the op indeed, which in my experience is useful when you get an error, but can't know from which promise it comes. Having a dedicated free of goja-ism function to try and handle that case is also a good idea, I'm gonna see what I can come up with. 

For the reference, i[n the initial PR, I had went through the hassle of implementing an "as close" as possible to the specs algorithm normalisation](https://github.com/grafana/xk6-webcrypto/pull/26/files#diff-6726d6a67040dc82c4a01ace327b184ce34fd868c3e16f17716b03bb3cd7a129R31). At the time, @olegbespalov argued that as long as the tests passed, it didn't matter if we implemented all the steps, especially considering that those were designed for IDL compliance; which we don't really care for in k6?

Both options are possible, and I went for a compromise 😄

oleiade:
@imiric:
- regarding the refactoring, I try to do so in general, but I must admit that I got myself the "feet stuck in the carpet" there, and for the sake of making the PRs consistent I kept the refactoring in as to not treat the refactoring out of context. I'll definitely do my best to make refactoring in separate PRs in the future indeed 🙇🏻 
- regarding writing keys, you're completely right, and it's something that I'd like [to address with better support for files in the close future](https://github.com/grafana/k6/issues/2977). In the meantime, users will only be able to `importKey` from the disk 👍🏻

webcrypto/algorithm.go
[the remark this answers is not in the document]
imiric: I don't mind the missing `op` handling. I agree with Oleg, and it's best to be pragmatic about this, as long as the tests pass.

But I do think we should fix the validation in `normalizeAlgorithm()` if the passed value is a string. It would be compliant with the spec, and less code overall.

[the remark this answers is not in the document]
oleiade: Fair enough @imiric. [I've updated the `normalizeAlgorithm` to be more accurate](https://github.com/grafana/xk6-webcrypto/pull/30/files#diff-261464395ae77be967b325f3b606881c51d29a3848d891e5bfe21561e4574df3R124), and to handle strings as-per-the-specs 🤞🏻 Let me know if that's satisfying 🙇🏻

oleiade:
As requested in #31 I added a README file with more information regarding the current state of supported operations and algorithms 🙇🏻

webcrypto/algorithm.go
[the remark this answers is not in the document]
imiric: LGTM :slightly_smiling_face: Now you don't need to reject the promise with `NotSupportedError` in `GenerateKey` or `Digest`, since the algorithm would've already been validated by `normalizeAlgorithm` in all cases. But it doesn't hurt leaving that in, I suppose.

README.md
```diff
@@ -1,3 +1,34 @@
...
+
+#### Crypto
+
+- `crypto.getRandomValues()`
+- `crypto.randomUUID()`
+
+#### SubtleCrypto
+
+- `crypto.subtle.digest()` supports the complete specification's algorithms list: SHA-1, SHA-256, SHA-384, SHA-512.
+- `crypto.subtle.generateKey()` supports the following symmetric algorithms: AES-CBC, AES-GCM, and AES-CTR.
```
codebien, line 19: I think a feature table could be more readable.

README.md
```diff
@@ -1,3 +1,34 @@
 # xk6-webcrypto
 
-This is a **work in progress** project implementation of the [WebCrypto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API) for k6.
\ No newline at end of file
+This is a **work in progress** project implementation of the [WebCrypto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API) specification for k6.
+
+## Current state
+
+The current state of the project is that it is an experimental, and incomplete implementation of the WebCrypto API specification. While we consider it ready for production use, it is still missing some features and is not yet fully compliant with the specification.
```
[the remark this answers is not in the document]
codebien: It sounds like a repetition, we already explicit it after.

mstoykov (approved):
LGTM!

I would prefer to merge PRs and fix random stylistic problems later, as we are definitely going to be rewrrting this quite a bit as we add more and more functionality. But we need something to build on top. 

And as was experienced - building everything in one go makes things even worse.

A stylistic note: In some places you use `switch` to check that a variable has one of many values, in some placse you use `isSomething` variables and and `if`.

For the majority of *this* particular cases I prefer the `switch` variant. It is both more concise and IMO is easier to read in almost all cases except the ones where for example we are counting `bits`. But even there it likely is good enough and we can have inline comment to help with that.

oleiade:
@mstoykov thanks for your feedback, I'm on your side with addressing stylistic issues separately. Regarding switch vs if/is I'm also on board. I'll keep in mind to be more consistent moving forward, and also expect that I will naturally rewrite most of the ifs as switch statements as more algorithms are supported 👍🏻

README.md
oleiade: You're right 👍🏻

codebien (approved):

PR #32, opened 2023-03-27
[3/3] Implement the `encrypt` and `decrypt` operations with support for AES algorithms

oleiade:
## What this is

This PR implements to support the [`encrypt`](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-encrypt) and [`decrypt`](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-decrypt)  operations.  To keep Pull Requests as small and manageable as possible, **It only implements support for the AES family of algorithms**.

⚠️ This PR is based on #31 and should only be merged once it is itself merged in `main` ✋🏻 

The [`encrypt` ](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-encrypt) operation allows encrypting a plaintext payload using an existing [`CryptoKey`](https://www.w3.org/TR/WebCryptoAPI/#cryptokey-interface) object. Note that although the common nomenclature refers to them as "text", those actually consist in `ArrayBuffer` and `TypedArray` in practice. The supported algorithms in this PR are only AES, and the supported AES ciphers are: CBC, CTR, and GCM as per the specification. Note that as a result of the Go standard library limitations, we had to bypass certain specification requirements: for instance, when using the GCM cipher, we only support iv sizes of 96bits, although the specs targets a much wider range. This is a result of the Go AES GCM implementation supporting 12 bytes nonces.

The [`decrypt`](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-decrypt) operation does the opposite. It allows decrypting a ciphertext into its original plaintext representation. Note that although the common nomenclature refers to them as "text", those actually consist in `ArrayBuffer` and `TypedArray` in practice. Same as with encryption, the supported algorithms in this PR are only AES, and the supported AES ciphers are: CBC, CTR, and GCM as per the specification. For the same reasons as with encryption, we are supporting only a subset of the specs in certain areas.

🟢 While the PR shows ~2000Loc changes, most of them are from the official test suites, which **do not need to be reviewed**, and the effective changes is more around ~600Loc 🟢 

### Test suite

While the PR might look big at first glance, you can safely disregard and ignore the files in the `webcrypto/tests` folder. Those files are directly imported and slightly adapted from [the official test suite](https://github.com/web-platform-tests/wpt/tree/master/WebCryptoAPI) so that they are runnable in k6, but we haven't written those originally. We won't make any functional changes to them, as they are our safety net, ensuring that our implementation matches the specifications expectation (and that what we do is 🥁 **safe**).

## References

This PR is the second chain of 3 implementing support for [key generation](https://github.com/grafana/xk6-webcrypto/pull/30) (1.), [import/export](https://github.com/grafana/xk6-webcrypto/pull/31#30) (2.), and encryption/decryption using them. We intend to merge these in order in master: 1->2->3.


webcrypto/subtle_crypto.go
```diff
@@ -35,9 +36,71 @@ type SubtleCrypto struct {
...
 //
 // The `data` parameter should contain the data to be encryption.
-func (sc *SubtleCrypto) Encrypt(algorithm goja.Value, key goja.Value, data []byte) *goja.Promise {
-	// TODO: implementation
-	return nil
+func (sc *SubtleCrypto) Encrypt(algorithm, key, data goja.Value) *goja.Promise {
+	rt := sc.vu.Runtime()
+	promise, resolve, reject := sc.makeHandledPromise()
+
+	// 2.
```
codebien, line 43:
```suggestion
```
Are they sequential numbers for the various operations? If yes, then I'm for removing them if we don't have a specific reason for keeping them.

webcrypto/subtle_crypto_test.go
```diff
@@ -92,3 +92,95 @@ func TestSubtleCryptoImportExport(t *testing.T) {
 		assert.NoError(t, gotScriptErr)
 	})
 }
+
+func TestSubtleCryptoEncryptDecrypt(t *testing.T) {
+	t.Parallel()
+
+	t.Run("AES CBC", func(t *testing.T) {
```
codebien, line 99: nit: I think we can put in dedicated funcs instead of nesting them.

codebien (changes requested):
>[Refactor CryptoKey to not use generics anymore](https://github.com/grafana/xk6-webcrypto/pull/32/commits/2446122efb3efb70348d3bcc36b7d9fbfa37b8e3)

@oleiade Why are we doing it in this PR and not refactoring directly from the PR it is adding?

webcrypto/subtle_crypto_test.go
oleiade: The benefit I see in doing it this way is that the test file, which is gonna end up being very big once all the algorithms and functionalities are implemented, remains nicely segmented. Another benefit is that it avoids longer test function names such as `TestSubtleCryptoEncryptDecryptAESGCMiv96bits`. If you don't oppose it, I'd rather keep it this way, let me know 😸

webcrypto/subtle_crypto.go
codebien: Let's say I'm a new maintainer of this library, how can I learn this convention without asking? Isn't a direct link to the spec a better solution for this problem?

oleiade: @codebien that's a fair point. I've added a section to the README detailing the convention, and providing some guidance as a result 👍🏻 🙇🏻 

Side note: in practice, I wouldn't expect maintainers or contributors to jump in without having the specs open, and it was my experience reviewing Deno's implementation that you immediately understand what those comments are for if you look at the specs. It also already proved useful in previous PRs where reviewers were able to spot a certain step happened to early, thanks to those comments 😄

codebien:
>unless your strategy was to preempt 3/3 to help me gain some time as there are other reviewers for the previous PRs already

Yeah, it was already covered by 2 reviewers, I started where it was uncovered.

webcrypto/subtle_crypto.go
```diff
@@ -60,9 +123,70 @@ func (sc *SubtleCrypto) Encrypt(algorithm goja.Value, key goja.Value, data []byt
...
+	}
+
+	var ck CryptoKey
+	if err = rt.ExportTo(key, &ck); err != nil {
+		reject(NewError(0, InvalidAccessError, "key argument does hold not a valid CryptoKey object"))
+		return promise
+	}
+
+	// 8.
+	keyAlgorithmName := key.ToObject(rt).Get("algorithm").ToObject(rt).Get("name").String()
```
mstoykov, line 152: This can panic if any of those `Get`'s return a null.

webcrypto/subtle_crypto.go
```diff
@@ -60,9 +123,70 @@ func (sc *SubtleCrypto) Encrypt(algorithm goja.Value, key goja.Value, data []byt
...
+			// 10.
+			plaintext, err = decrypter.Decrypt(ciphertext, ck)
+			if err != nil {
+				reject(err)
+				return
+			}
+
+			resolve(rt.NewArrayBuffer(plaintext))
+		default:
+			reject(NewError(0, OperationError, fmt.Sprintf("unsupported algorithm %q", normalized.Name)))
```
codebien, line 184:
```suggestion
			reject(NewError(0, UnsupportedError, fmt.Sprintf("unsupported algorithm %q", normalized.Name)))
```
Should it be an unsupported error?

webcrypto/aes.go
```diff
@@ -151,3 +155,458 @@ func importAESKey(
...
+
+var (
+	// Ensure that AesCtrParams implements the Encrypter interface.
+	_ Encrypter = &AesCtrParams{}
+
+	// Ensure that AesCtrParams implements the Decrypter interface.
+	_ Decrypter = &AesCtrParams{}
+
+	// Ensure that AesCtrParams implements the EncryptDecrypter interface.
+	_ EncryptDecrypter = &AesCtrParams{}
```
codebien, line 368:
```suggestion
	// Ensure that AesCtrParams implements the EncryptDecrypter interface.
	_ EncryptDecrypter = &AesCtrParams{}
```
EncryptDecrypter should be enough for all of them

webcrypto/aes.go
```diff
@@ -151,3 +155,458 @@ func importAESKey(
...
+
+	// 2.
+	// As described in section 8 of AES-GCM [NIST SP800-38D].
+	// [NIST SP800-38D] https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38d.pdf
+	//
+	// NOTE @oleiade: that the spec targets to support multiple IV lengths
+	// but go only supports 12 bytes IVs. We therefore are diverging from the
+	// spec here, and have adjusted the test suite accordingly.
+	if len(agp.Iv) != 12 {
+		return nil, NewError(0, OperationError, "iv length is not 12 bytes long")
```
codebien, line 439:
```suggestion
		return nil, NewError(0, NotSupportedErrror, "only 12 bytes long iv length are supported")
```
I hope it helps to understand better why

webcrypto/aes.go
```diff
@@ -151,3 +155,458 @@ func importAESKey(
...
+
+	gcm, err := cipher.NewGCMWithTagSize(block, tagLength/8)
+	if err != nil {
+		return nil, NewError(0, ImplementationError, "could not create cipher")
+	}
+
+	// The Golang AES GCM cipher only supports a Nonce/Iv length of 12 bytes,
+	// as opposed to the looser requirements of the Web Crypto API spec.
+	if len(agp.Iv) != gcm.NonceSize() {
+		return nil, NewError(0, OperationError, "iv length is not 12")
```
codebien, line 484: same

webcrypto/subtle_crypto.go
```diff
@@ -60,9 +123,70 @@ func (sc *SubtleCrypto) Encrypt(algorithm goja.Value, key goja.Value, data []byt
...
+			return
```
mstoykov, line 185:
```suggestion
```

webcrypto/subtle_crypto.go
```diff
@@ -35,9 +36,71 @@ type SubtleCrypto struct {
...
+		}
+
+		var ciphertext []byte
+
+		switch normalized.Name {
+		case AESCbc, AESCtr, AESGcm:
+			// 10.
+			ciphertext, err = encrypter.Encrypt(plaintext, ck)
+			if err != nil {
+				reject(err)
```
codebien, line 91: Is an error generated here, clear enough or should we wrap it?

webcrypto/aes.go
```diff
@@ -151,3 +155,458 @@ func importAESKey(
...
+//
+// [specification]: https://www.w3.org/TR/WebCryptoAPI/#aes-ctr
+func (acp *AesCtrParams) Decrypt(ciphertext []byte, key CryptoKey) ([]byte, error) {
+	// 1.
+	if len(acp.Counter) != aes.BlockSize {
+		return nil, NewError(0, OperationError, "counter length is invalid, should be 16 bytes")
+	}
+
+	// 2.
+	if acp.Length == 0 || acp.Length > 128 {
```
mstoykov, line 338:
```suggestion
	if acp.Length <= 0 || acp.Length > 128 {
```

webcrypto/encryption.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+
+		ed = &ctrParams
+	case AESGcm:
+		var gcmParams AesGcmParams
+		if err = rt.ExportTo(params, &gcmParams); err != nil {
+			return nil, NewError(0, SyntaxError, "invalid algorithm parameters, unable to interpret as AesGcmParams object")
+		}
+
+		ed = &gcmParams
```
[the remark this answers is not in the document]
oleiade: I've tried a few things to dry it, but not that yet! Good intuition, I'll try it 👍🏻

webcrypto/subtle_crypto.go
oleiade: I have been careful to do so indeed, and double checked, in every cases, the Encrypter will return a rich error 👍🏻 The specification does not expect the operation to fail, so we have free hands there. I'd prefer to keep this granularity as it would help debugging user issues if they occur. Feel free to unresolve if you would prefer it another way 🙇🏻

webcrypto/subtle_crypto.go
```diff
@@ -60,9 +123,70 @@ func (sc *SubtleCrypto) Encrypt(algorithm goja.Value, key goja.Value, data []byt
...
+		}
+
+		var plaintext []byte
+
+		switch normalized.Name {
+		case AESCbc, AESCtr, AESGcm:
+			// 10.
+			plaintext, err = decrypter.Decrypt(ciphertext, ck)
+			if err != nil {
+				reject(err)
```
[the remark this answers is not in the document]
oleiade: I have been careful to do so indeed, and double checked, in every cases, the Encrypter will return a rich error 👍🏻 The specification does not expect the operation to fail, so we have free hands there. I'd prefer to keep this granularity as it would help debugging user issues if they occur. Feel free to unresolve if you would prefer it another way 🙇🏻

webcrypto/aes.go
```diff
@@ -151,3 +155,458 @@ func importAESKey(
...
+type AesCbcParams struct {
+	Algorithm
+
+	// Name should be set to AES-CBC.
+	Name string `json:"name"`
+
+	// Iv holds (an ArrayBuffer, a TypedArray, or a DataView) the initialization vector.
+	// Must be 16 bytes, unpredictable, and preferably cryptographically random.
+	// However, it need not be secret (for example, it may be transmitted unencrypted along with the ciphertext).
+	Iv []byte `json:"iv"`
```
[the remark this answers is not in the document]
oleiade: In theory I would agree with you, but in practice, the aes cipher functions expect a slice, thus I would need to convert it to a slice with the `[:]` which, I believe would lead to effectively copy the content of the array on the stack on the heap: which would be inefficient? I might have the wrong assumption though.

Also, quickly testing the solution, it seems Goja's `ExportTo` method's not fond of arrays.

webcrypto/subtle_crypto.go
oleiade: I've added a `traverseObject` helper to handle that case, and used it in `Encrypt`/`Decrypt` 👍🏻

webcrypto/subtle_crypto.go
```diff
@@ -60,9 +123,70 @@ func (sc *SubtleCrypto) Encrypt(algorithm goja.Value, key goja.Value, data []byt
 // The `key` parameter should be a `CryptoKey` to be used for decryption.
 //
 // The `data` parameter should contain the data to be decrypted.
-func (sc *SubtleCrypto) Decrypt(algorithm goja.Value, key goja.Value, data []byte) *goja.Promise {
-	// TODO: implementation
-	return nil
+func (sc *SubtleCrypto) Decrypt(algorithm, key, data goja.Value) *goja.Promise {
```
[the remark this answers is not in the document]
oleiade: I've been experimenting around this idea, and landed on a pretty promising idea indeed. However, I'd prefer to wait until the rest of the algorithms (especially RSA) are implemented to try and DRY it. That will help maintain the cognitive load introduced by abstractions and special cases handling low. 

I've opened an [issue](#34 ) to not lose track of this. Feel free to unresolve if you would prefer to address it in this PR specifically 👍🏻 🙇🏻

codebien (approved):

webcrypto/goja.go
```diff
@@ -29,6 +33,34 @@ func exportArrayBuffer(rt *goja.Runtime, v goja.Value) ([]byte, error) {
 	return ab.Bytes(), nil
 }
 
+// traverseObject traverses the given object using the given fields and returns the value
+// at the end of the traversal. It assumes that all the traversed fields are Objects.
+func traverseObject(rt *goja.Runtime, src *goja.Object, fields ...string) (goja.Value, error) {
+	var value goja.Value = src
```
mstoykov, line 39:
```suggestion
func traverseObject(rt *goja.Runtime, value goja.Value, fields ...string) (goja.Value, error) {
```
I was going to recommend this at the call site, but there is no need for you to make the value into an object if all of that can be done here to begin with.

webcrypto/goja.go
```diff
@@ -29,6 +33,34 @@ func exportArrayBuffer(rt *goja.Runtime, v goja.Value) ([]byte, error) {
...
+	var currentField string
+	var obj *goja.Object
+
+	for idx, field := range fields {
+		if idx > 0 {
+			currentField += "."
+		}
+
+		currentField += field
```
mstoykov, line 48: This adds quite a bit of allocations for an error message that *hopefully* will never happen. 

This should go "backwards" and build the message when needed.

I am pretty sure `strings.Join(fields[:idx], ",")` has the same end result and can be calculated only once, and only when needed.

webcrypto/goja.go
```diff
@@ -29,6 +33,34 @@ func exportArrayBuffer(rt *goja.Runtime, v goja.Value) ([]byte, error) {
...
+
+		obj = value.ToObject(rt)
+		if isNullish(obj) {
```
mstoykov, line 51: I don't think this is possible at all - I am pretty sure `ToObject` panics in those cases, and we do check the value after we get it in the line below.

It might be worth moving this to above the `for` for the input `value` especially if we refactor it to be taking a value instead of an object.

mstoykov (approved):
I have comments around the new `traverseObject`, but we can move them in a new issue if you want.

PR #38, opened 2023-03-30
[2/3] Implement the `exportKey` and `importKey` operations with support for HMAC algorithm

oleiade:
## What is this?

This Pull Request adds support for the HMAC algorithm to the [`subtle.crypto.importKey`](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-importKey) and [`subtle.crypto.exportKey`](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-exportKey) operations. It is the second PR of a chain of three PRs converging towards introducing the `sign` and `verify` APIs to the extension. 

Our plan moving forward is to go breadth first, and implement as much of the WebCrypto API's surface, and only later to add support for more algorithms. As a result this PR comes in with the intention to support the ability to implement the [`sign`](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-sign) and [`verify`](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-verify) operations to the module.

Just like with `encrypt` and `decrypt`, the sign and verify operations test suite rely on the ability to import and export a key of a support algorithm. As a result to support sign and verify with a scope reduced to the HMAC algorithm, we need to implement `importKey` and `exportKey` first.

## Review

Note that like in the `generateKey` PR we introduce a new interface for key import: `KeyImporter`. Although we could have extended this principle to key export, we decided to delay that until we support key pairs, as we lack visibility to decide whether or not this will remain a good abstraction in the future.

This PR is based on the [HMAC support for the `generateKey`](https://github.com/grafana/xk6-webcrypto/pull/35) PR, and will act as the base for a third PR to come which will introduce the `sign` and `verify` methods to the extension.

### DIFF
--- a/webcrypto/hmac.go
+++ b/webcrypto/hmac.go
@@ -185,3 +185,135 @@ func exportHmacKey(ck *CryptoKey, format KeyFormat) ([]byte, error) {
 		return nil, NewError(0, NotSupportedError, "unsupported key format "+format)
 	}
 }
+
+// HmacImportParams represents the object that should be passed as the algorithm parameter
+// into `SubtleCrypto.GenerateKey`, when generating an HMAC key.
+type HmacImportParams struct {
+	Algorithm
+
+	// Hash represents the name of the digest function to use. You can
+	// use any of the following: [Sha256], [Sha384],
+	// or [Sha512].
+	Hash Algorithm `json:"hash"`
+
+	// Length holds (a Number) the length of the key, in bits.
+	// If this is omitted, the length of the key is equal to the block size
+	// of the hash function you have chosen.
+	// Unless you have a good reason to use a different length, omit
+	// use the default.
+	Length null.Int `json:"length"`
+}
+
+// newHmacImportParams creates a new HmacImportParams object from the given
+// algorithm and params objects.
+func newHmacImportParams(rt *goja.Runtime, normalized Algorithm, params goja.Value) (*HmacImportParams, error) {
+	// The specification doesn't explicitly tell us what to do if the
+	// hash field is not present, but we assume it's a mandatory field
+	// and throw an error if it's not present.
+	hashValue, err := traverseObject(rt, params, "hash")
+	if err != nil {
+		return nil, NewError(0, SyntaxError, "could not get hash from algorithm parameter")
+	}
+
+	// Although the specification doesn't explicitly ask us to do so, we
+	// normalize the hash algorithm here, as it shares the same definition
+	// as the AlgorithmIdentifier type, and we'll need it later on.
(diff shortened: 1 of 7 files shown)


webcrypto/hmac.go
```diff
@@ -185,3 +185,135 @@ func exportHMACKey(ck *CryptoKey, format KeyFormat) ([]byte, error) {
...
+	// 3.
+	var hash KeyAlgorithm
+
+	// 4.
+	switch format {
+	case RawKeyFormat:
+		hash = KeyAlgorithm{Algorithm{Name: hip.Hash.Name}}
+	default:
+		return nil, NewError(0, NotSupportedError, "unsupported key format "+format)
+	}
```
codebien, line 273: Should we centralize this logic in the key algorithm?

webcrypto/subtle_crypto.go
```diff
@@ -520,29 +521,38 @@ func (sc *SubtleCrypto) ImportKey(
...
-			if (result.Type == SecretCryptoKeyType || result.Type == PrivateCryptoKeyType) && len(keyUsages) == 0 {
-				reject(NewError(0, SyntaxError, "the keyUsages argument must contain at least one valid usage for the algorithm"))
-				return
-			}
+		// 9.
+		isSecretKey := result.Type == SecretCryptoKeyType
+		isPrivateKey := result.Type == PrivateCryptoKeyType
+		isUsagesEmpty := len(keyUsages) == 0
+		if (isSecretKey || isPrivateKey) && isUsagesEmpty {
+			reject(NewError(0, SyntaxError, "usages cannot not be empty for a secret or private CryptoKey"))
```
codebien, line 544: Is syntax the proper error here?

codebien (commented):
LGTM

webcrypto/subtle_crypto.go
oleiade: It is indeed, as detailed in step `9.` of the `[importKey](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-importKey)` algorithm 👍🏻

webcrypto/hmac.go
oleiade: I'm not sure I understand what you mean, could you elaborate please? 😄

codebien: I had the impression this was the same logic applied across different keys but it is not true. The switch is the same but the logic for the case is different. You can skip this.

codebien (approved):
LGTM

mstoykov (approved):

PR #40, opened 2023-04-03
[3/3] Implement the `sign` and `verify` operations with support for HMAC algorithm

oleiade:
## What is this?

This Pull Request adds support for the HMAC algorithm to the [`sign`](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-sign) and [`verify`](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-verify) operations.

## Review

This PR is based on the [HMAC support for the `importKey` and `exportKey`](https://github.com/grafana/xk6-webcrypto/pull/38) PR.

You might also notice that some tests have been commented out. Those are momentarily deactivated as support for [ECDSA key generation is a requirement](#41). We do not support that algorithm for key generation yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,8 +29,8 @@ The current state of the project is that it is an experimental module of the Web
 
 | API                      | HMAC | ECDSA | RSASSA-PKCS1-v1_5 | RSA-PSS |
 | :----------------------- | :--- | :---- | :---------------- | :------ |
-| `crypto.subtle.sign()`   | ❌   | ❌    | ❌                | ❌      |
-| `crypto.subtle.verify()` | ❌   | ❌    | ❌                | ❌      |
+| `crypto.subtle.sign()`   | ✅   | ❌    | ❌                | ❌      |
+| `crypto.subtle.verify()` | ✅   | ❌    | ❌                | ❌      |
 
 ##### Key generation, import and export
 


webcrypto/subtle_crypto.go
```diff
@@ -218,9 +217,86 @@ func (sc *SubtleCrypto) Decrypt(algorithm, key, data goja.Value) *goja.Promise {
...
+
+	go func() {
+		// 8.
+		if normalized.Name != keyAlgorithmNameValue.String() {
+			reject(NewError(0, InvalidAccessError, "algorithm name does not match key algorithm name"))
+			return
+		}
+
+		// 9.
+		for !containsUsage(ck.Usages, SignCryptoKeyUsage) {
```
codebien, line 261: Should we move `containsUsage` function on the crypto key?
```suggestion
		for !ck.containsUsage(SignCryptoKeyUsage) {
```

webcrypto/subtle_crypto.go
```diff
@@ -246,14 +322,92 @@ func (sc *SubtleCrypto) Sign(algorithm goja.Value, key goja.Value, data []byte)
...
+	promise, resolve, reject := sc.makeHandledPromise()
+
+	// 2.
+	ab, err := exportArrayBuffer(sc.vu.Runtime(), signature)
+	if err != nil {
+		reject(err)
+		return promise
+	}
+	signatureData := make([]byte, len(ab))
+	copy(signatureData, ab)
```
codebien, line 338: Should we wrap it a function? Like `exportAndCopyArrayBuffer()`

codebien (changes requested):

webcrypto/subtle_crypto.go
oleiade: I considered it, but I tend to prefer composition over packing everything in a single function 😄 In the Web Platform tests they have this `copyBuffers` helper function everywhere, I could add a similar function in Go `copyBuffers(src, dst goja.ArrayBuffer), what do you think?

webcrypto/subtle_crypto.go
```diff
@@ -218,9 +217,86 @@ func (sc *SubtleCrypto) Decrypt(algorithm, key, data goja.Value) *goja.Promise {
...
+				return
+			}
+
+			hashFn, ok := getHashFn(keyAlgorithm.Hash.Name)
+			if !ok {
+				reject(NewError(0, NotSupportedError, fmt.Sprintf("unsupported key hash algorithm %q", keyAlgorithm.Hash.Name)))
+				return
+			}
+
+			hasher := hmac.New(hashFn, keyHandle)
```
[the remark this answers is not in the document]
oleiade: We also use that function in the `digest` operation, so I would probably keep it as a helper function. But I do like the idea of having a `HashFn()` method on the `KeyAlgorithm` struct though, what do you think?

webcrypto/subtle_crypto.go
oleiade: Considering we only use `exportArrayBuffer` in the context of steps described by the specification as "obtain a copy of the bytes held by X", I modified the function to return a copy of the underlying bytes 👍🏻

codebien (approved):
:clap:

PR #45, opened 2023-04-14
Refactor bits/bytes conversion

### LINKED ISSUE #36: I am not really happy with this `*8` and then `/8` in places.
I am not really happy with this `*8` and then `/8` in places.

It feels to me like we can have a small `type bitLength int` (or byteLength) and it haviing `bitsLength()` and `bytesLength()` and a constructor that take care of that under the hood.

I feel like in a lot of places we do operations like this in this module, and we will get one of the wrong and will find out much, much later. 

Not to speak it is kind of strange we do it withoug going to read the `Length` is in bits instead of `bytes`.

_Originally posted by @mstoykov in https://github.com/grafana/xk6-webcrypto/pull/35#discussion_r1153151192_

oleiade:
This PR address #36 and introduces two new type aliases: `bitLength` and `byteLength` helping make conversion from bytes length to bits and vice versa, more explicit, readable, and less error prone.

closes #36

### DIFF
--- a/webcrypto/aes.go
+++ b/webcrypto/aes.go
@@ -20,7 +20,7 @@ type AESKeyGenParams struct {
 	Algorithm
 
 	// The length, in bits, of the key.
-	Length int64 `json:"length"`
+	Length bitLength `json:"length"`
 }
 
 // newAESKeyGenParams creates a new AESKeyGenParams object, from the
@@ -41,7 +41,7 @@ func newAESKeyGenParams(rt *goja.Runtime, normalized Algorithm, params goja.Valu
 
 	return &AESKeyGenParams{
 		Algorithm: normalized,
-		Length:    algorithmLength,
+		Length:    bitLength(algorithmLength),
 	}, nil
 }
 
@@ -75,7 +75,7 @@ func (akgp *AESKeyGenParams) GenerateKey(
 		return nil, NewError(OperationError, "invalid key length")
 	}
 
-	randomKey := make([]byte, akgp.Length/8)
+	randomKey := make([]byte, akgp.Length.asByteLength())
 	if _, err := rand.Read(randomKey); err != nil {
 		// 4.
 		return nil, NewError(OperationError, "could not generate random key")
@@ -86,7 +86,7 @@ func (akgp *AESKeyGenParams) GenerateKey(
 	key.Type = SecretCryptoKeyType
 	key.Algorithm = AESKeyAlgorithm{
 		Algorithm: akgp.Algorithm,
-		Length:    akgp.Length,
+		Length:    int64(akgp.Length),
 	}
 
 	// 10.
@@ -190,7 +190,7 @@ func (aip *AESImportParams) ImportKey(
 	key := &CryptoKey{
 		Algorithm: AESKeyAlgorithm{
 			Algorithm: aip.Algorithm,
-			Length:    int64(len(keyData) * 8),
+			Length:    int64(byteLength(len(keyData)).asBitLength()),
 		},
 		Type:   SecretCryptoKeyType,
 		handle: keyData,
@@ -443,7 +443,7 @@ type AESGCMParams struct {
 	// in some applications: Appendix C of the specification provides additional guidance here.
 	//
 	// tagLength is optional and defaults to 128 if it is not specified.
-	TagLength int `json:"tagLength"`
+	TagLength bitLength `json:"tagLength"`
 }
 
 // Encrypt encrypts the given plaintext using the AES-GCM algorithm, and returns the ciphertext.
@@ -477,7 +477,7 @@ func (agp *AESGCMParams) Encrypt(plaintext []byte, key CryptoKey) ([]byte, error
 	}
 
 	// 4.
-	var tagLength int
+	var tagLength bitLength
 	if agp.TagLength == 0 {
 		tagLength = 128
 	} else {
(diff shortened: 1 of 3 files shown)


webcrypto/aes.go
```diff
@@ -75,7 +75,7 @@ func (akgp *AesKeyGenParams) GenerateKey(
 		return nil, NewError(0, OperationError, "invalid key length")
 	}
 
-	randomKey := make([]byte, akgp.Length/8)
+	randomKey := make([]byte, bitLength(akgp.Length).asByteLength())
```
imiric, line 78: Have you considered making `AesKeyGenParams.Length` of type `bitLength` to avoid the conversion?

imiric (approved):

webcrypto/aes.go
oleiade: I was a bit on the fence about it; but as you bring it up, I'll give it a shot, let's see if Goja handles that well 🤞🏻

oleiade: Goja was happy 😀 Applied the change indeed, thanks a lot 👍🏻 🙇‍♂️

webcrypto/types.go
```diff
@@ -6,6 +6,22 @@ import (
 	"github.com/dop251/goja"
 )
 
+// bitLength is a type alias for the length of a bits collection.
+type bitLength int
+
+// asByteLength returns the length of the bits collection in bytes.
+func (b bitLength) asByteLength() int {
```
mstoykov, line 13: wouldn't it be better if this actually returns `byteLength`?

I did try locally and while this means that now at some places we need to cast it to `int` int also removed the need to cast it to byteLength in kind of the same amount of places.

Same for `byteLength`.

oleiade: I tried that too, but we do have to make a bunch casts as a result indeed, which felt a bit like it was defeating the purpose, and I was waiting to see what you thought 👍🏻 

I'll go ahead and change the return types then 🙇🏻

webcrypto/aes.go
```diff
@@ -544,7 +544,7 @@ func (agp *AESGCMParams) Decrypt(ciphertext []byte, key CryptoKey) ([]byte, erro
 	// 2.
 	// Note that we multiply the length of the ciphertext by 8, in order
 	// to get the length in bits.
-	if len(ciphertext)*8 < tagLength {
+	if int(byteLength(len(ciphertext)).asBitLength()) < tagLength {
```
mstoykov, line 547: why not make `tagLength` a `bitLength` instead?

webcrypto/aes.go
```diff
@@ -86,7 +86,7 @@ func (akgp *AESKeyGenParams) GenerateKey(
 	key.Type = SecretCryptoKeyType
 	key.Algorithm = AESKeyAlgorithm{
 		Algorithm: akgp.Algorithm,
-		Length:    akgp.Length,
+		Length:    int64(akgp.Length),
```
mstoykov, line 89: Can we change that `Length` field  as well?

webcrypto/hmac.go
```diff
@@ -285,28 +285,28 @@ func (hip *HMACImportParams) ImportKey(
...
 			return nil, NewError(DataError, "key length cannot be greater than the length of the imported data")
 		}
 
-		if hip.Length.Int64 < length {
+		if hip.Length.Int64 < int64(length) {
 			return nil, NewError(DataError, "key length cannot be less than the length of the imported data")
 		}
 
-		length = hip.Length.Int64
+		length = bitLength(hip.Length.Int64)
```
mstoykov, line 303: Looking at this code:
1. do we need the last assignement? If it is not less and it is not more ... and it is an int* isn't it already equal?
2. Can we maybe make this just check that they are not equal - I don't think that it matters all that much to tell the user whether it is smaller or bigger.

webcrypto/hmac.go
```diff
@@ -100,13 +100,13 @@ func (hkgp *HMACKeyGenParams) GenerateKey(
...
+			hkgp.Length = null.IntFrom(int64(byteLength(sha1.BlockSize).asBitLength()))
 		case SHA256:
-			hkgp.Length = null.IntFrom(sha256.BlockSize * 8)
+			hkgp.Length = null.IntFrom(int64(byteLength(sha256.BlockSize).asBitLength()))
 		case SHA384:
-			hkgp.Length = null.IntFrom(sha512.BlockSize * 8)
+			hkgp.Length = null.IntFrom(int64(byteLength(sha512.BlockSize).asBitLength()))
 		case SHA512:
-			hkgp.Length = null.IntFrom(sha512.BlockSize * 8)
+			hkgp.Length = null.IntFrom(int64(byteLength(sha512.BlockSize).asBitLength()))
```
mstoykov, line 109: This will be *slightly* easier if we had `nullBitLength` but I don't know if it is worth doing in practice.

I would though recommend having a variable that is set in the cases and moving the actual `null.IntFrom` after the select.

mstoykov (approved):

webcrypto/aes.go
oleiade: Good call, I completely overlooked it :+1:

webcrypto/hmac.go
oleiade: I strictly followed the steps of the [specification](https://www.w3.org/TR/WebCryptoAPI/#hmac-operations) there. But you're right in the end it doesn't make a big difference in that case. I've reorganized the code accordingly :+1:.

webcrypto/hmac.go
oleiade: Funnily enough, I had started implementing `nullBitLength`, but quickly reconsidered as it was a lot of code introduced for a minimal use case at this point. We might reconsider at a later point in time if this pattern reemerges :+1:

mstoykov (approved):

PR #46, opened 2023-04-14
Refactor naming to follow Go naming conventions

### LINKED ISSUE #39: Make naming consistent throughout the codebase
Although Go advocates for a naming convention which keeps acronyms uppercased, the WebCrypto API doesn't follow such convention. It leads to certain types/constructs having names such as `AesKeyGenParams` when in proper Go code, it would be expected to be `AESKeyGenParams`.

So far we have followed the specs convention for structs, and Go's in other places such as for constants.

We should pick one, stick to it, and document it to make maintainers' lives easier 👍🏻

oleiade:
This PR renames most structs and constants involving an acronym or cipher name following the Go convention of holding to acronyms as capitalised strings.

closes #39

### DIFF
--- a/webcrypto/aes.go
+++ b/webcrypto/aes.go
@@ -10,25 +10,25 @@ import (
 	"github.com/dop251/goja"
 )
 
-// AesKeyGenParams represents the object that should be passed as
+// AESKeyGenParams represents the object that should be passed as
 // the algorithm parameter into `SubtleCrypto.generateKey`, when generating
 // an AES key: that is, when the algorithm is identified as any
 // of AES-CBC, AES-CTR, AES-GCM, or AES-KW.
 //
 // [specification]: https://www.w3.org/TR/WebCryptoAPI/#aes-keygen-params
-type AesKeyGenParams struct {
+type AESKeyGenParams struct {
 	Algorithm
 
 	// The length, in bits, of the key.
 	Length int64 `json:"length"`
 }
 
-// newAesKeyGenParams creates a new AesKeyGenParams object, from the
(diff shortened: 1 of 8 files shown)


webcrypto/aes.go
```diff
@@ -395,16 +395,16 @@ func (acp *AesCtrParams) Decrypt(ciphertext []byte, key CryptoKey) ([]byte, erro
...
 
-// AesGcmParams represents the object that should be passed as the algorithm parameter
+// AESGcmParams represents the object that should be passed as the algorithm parameter
 // into `SubtleCrypto.Encrypt`, `SubtleCrypto.Decrypt`, `SubtleCrypto.WrapKey`, or
 // `SubtleCrypto.UnwrapKey`, when using the AES-GCM algorithm.
 // As defined in the [specification].
 //
 // [specification]: https://www.w3.org/TR/WebCryptoAPI/#aes-gcm-params
-type AesGcmParams struct {
+type AESGcmParams struct {
```
codebien, line 407:
```suggestion
type AESGCMParams struct {
```
It is an acronym so it should be all uppercase. I can see the readability issue. I'm not against the current version, I see this suggestion has a readability issue. `AES_GCMParams` should also works.

What is your idea here?

codebien (commented):
LGTM, just a comment regarding the combined acronyms.

webcrypto/aes.go
oleiade: You're right, GCM stands for Galois Counter Mode, and is an acronym, we should probably uppercase it indeed, I just didn't pay enough attention ✋🏻

Issue #57, opened 2024-01-12
Use modulestest for test


mstoykov:
modulestest tries to make a close as possible runtime so that when module test are written they actually represent as closely as possible the real k6.

In a recent case #55 and the it's fix #56 would likely have been much easier to spot and do - or not happen at all.

PR #59, opened 2024-01-12
Make newTestSetup rely on modulestest

oleiade:
This PR addresses #57 and modifies the `newTestSetup` helper to use `modulestest.NewRuntime` under the hood. This simplifies the test setup greatly, and delegates parts of the runtime setup to the core module.

⚠️ we should merge this after #58 to benefit from the linter issues fix ⚠️


webcrypto/test_setup.go
```diff
@@ -54,51 +35,14 @@ func newTestSetup(t testing.TB) testSetup {
...
-
-	vu := &modulestest.VU{
-		CtxField:     tb.Context,
-		InitEnvField: &common.InitEnvironment{},
-		RuntimeField: rt,
-		StateField:   state,
-	}
+	m := new(RootModule).NewModuleInstance(ts.VU)
+	require.NoError(t, ts.VU.Runtime().Set("crypto", m.Exports().Named["crypto"]))
+	require.NoError(t, ts.VU.Runtime().GlobalObject().Set("CryptoKey", CryptoKey{}))
```
mstoykov, line 43: you might want to make it also use the module system and then use a simple script to make it globally accessible as in 

https://github.com/grafana/k6/blob/f38347d45bceb8bb42bfbd8785d1d73940aa8b6b/js/modules/k6/experimental/fs/module_test.go#L684-L706

webcrypto/test_setup.go
```diff
@@ -9,34 +9,15 @@ import (
 
 	"github.com/dop251/goja"
 	"github.com/stretchr/testify/require"
-	"go.k6.io/k6/js/common"
-	"go.k6.io/k6/js/eventloop"
 	"go.k6.io/k6/js/modulestest"
-	"go.k6.io/k6/lib"
-	"go.k6.io/k6/lib/testutils/httpmultibin"
-	"go.k6.io/k6/metrics"
-	"gopkg.in/guregu/null.v3"
```
mstoykov, line 18: Dropping this dependancies is breaking the ci - running `go mod tidy` will fix it.

webcrypto/crypto_test.go
```diff
@@ -10,10 +11,10 @@ import (
 func TestGetRandomValues(t *testing.T) {
 	t.Parallel()
 
-	ts := newTestSetup(t)
+	ts, err := newConfiguredRuntime(t)
+	require.NoError(t, err)
 
-	gotScriptErr := ts.EventLoop.Start(func() error {
-		_, err := ts.VU.Runtime().RunString(`
+	_, gotScriptErr := ts.RunOnEventLoop(`
```
mstoykov, line 17: nit: This name threw me for a spin. I thought it was a boolean.

Looking at the previous code I guess this was so that the internal `err` isn't shadowing this one?

This is no longer teh case so maybe lets rename it ?

webcrypto/crypto_test.go
```diff
@@ -10,10 +11,10 @@ import (
 func TestGetRandomValues(t *testing.T) {
 	t.Parallel()
 
-	ts := newTestSetup(t)
+	ts, err := newConfiguredRuntime(t)
+	require.NoError(t, err)
```
mstoykov, line 15: nit: probably better to just `require.NoError` in the `newConfiguredRuntime` - I don't think there is a case for it returning error and us not stopping a test

mstoykov (approved):
LGTM! 


Left some nits if there is time

webcrypto/crypto_test.go
oleiade: You're right 👍🏻 I used the previous name, but with the `RunOnEventLoop` it doesn't feel fitting anymore, gonna change it 👏🏻

PR #60, opened 2024-01-19
Leverage some of the k6 APIs to handle Javascript operations

oleiade:
This is a small relatively small PR bringing two improvements to the webcrypto extension, and leveraging newer k6 versions:
- Use `common.IsNullish` instead of doing it ourselves
- Use `promises.New`

Let me know if you spot further improvements that could make sense in this context 🙇🏻


joanlopez (approved):

codebien (approved):
:rocket:

Issue #37, opened 2023-03-30
Implement support for the JWK format


oleiade:
The [Webcrypto API ](https://www.w3.org/TR/WebCryptoAPI/) specification has support for the [JWK](https://www.rfc-editor.org/rfc/rfc7517) key serialisation format. However, it is currently not implemented in this extension for a collection of reasons. If those were to be addressed, the implementation of the JWK format would therefore become unblocked.

## Blocker Reasons

- [ ] JWK has no official support in the Go standard library: we see it as a security issue not having an officially supported and reviewed implementation available.
- [ ] The test suite depending on JWK, such as import/export depend on the [`btoa` WebAPI function](https://developer.mozilla.org/en-US/docs/Web/API/btoa), which is not available in Goja. We could circumvent this issue by implementing an alternative ourselves, but it would take us further from being as close as possible from the original [WebPlatform test suite](https://github.com/web-platform-tests/wpt/tree/master/WebCryptoAPI).

PR #61, opened 2024-02-22
JWK Implementation for import/export

olegbespalov:
# What?

This is an implementation of the JWK export/import.

I've decided to use the popular library instead of self-tailored since, for the following algorithms, the JWK could be more complicated in terms of maintaining standards. 

### How did I test it?

* I did rely on the existing Web Platform tests (uncommented cases).
* also I've checked that the generated in nodejs key could be imported by the xk6-webcrypto

### Related PRs:

This PR is based on #62 

# Why?

JWK is part of the specification.

Closes: #37

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,8 +37,11 @@ The current state of the project is that it is an experimental module of the Web
 | API                           | AES-CBC | AES-GCM | AES-CTR | AES-KW | HMAC | ECDSA | ECDH | RSASSA-PKCS1-v1_5 | RSA-PSS | RSA-OAEP |
 | :---------------------------- | :------ | :------ | :------ | :----- | :--- | :---- | :--- | :---------------- | :------ | :------- |
 | `crypto.subtle.generateKey()` | ✅      | ✅      | ✅      | ❌     | ✅   | ❌    | ❌   | ❌                | ❌      | ❌       |
-| `crypto.subtle.importKey()`   | ✅      | ✅      | ✅      | ❌     | ❌   | ❌    | ❌   | ❌                | ❌      | ❌       |
-| `crypto.subtle.exportKey()`   | ✅      | ✅      | ✅      | ❌     | ❌   | ❌    | ❌   | ❌                | ❌      | ❌       |
+| `crypto.subtle.importKey()`   | ✅      | ✅      | ✅      | ❌     | ✅   | ❌    | ❌   | ❌                | ❌      | ❌       |
+| `crypto.subtle.exportKey()`   | ✅      | ✅      | ✅      | ❌     | ✅   | ❌    | ❌   | ❌                | ❌      | ❌       |
+
+> [!WARNING]  
+> Currently, only the `raw` and `jwk` (JSON Web Key) formats are supported for import/export operations.
 
 ##### Key derivation
 


oleiade (approved):
Great job @olegbespalov 👏🏻 LGTM 🚀 🙇🏻

oleiade:
Before I forget, we should also update the README tables to match the newly introduced support for JWK in `import` and `export` operations. At least for the sake of future selves 😸

olegbespalov:
@oleiade, the readme tables have no JWK dimension, but you are correct that it's worth mentioning the current state. I've made a note in [ea6ac87](https://github.com/grafana/xk6-webcrypto/pull/61/commits/ea6ac873a81be4ce1e64109bc373df915414b84d)

examples/import_export/import-export-jwk-aes-static-key.js
```diff
@@ -0,0 +1,31 @@
+import { crypto } from "k6/x/webcrypto";
+
+ export default async function () {
+    try {
```
joanlopez, line 4: Stupid question; why do this example is surrounded by a `try..catch`, while other (e.g. `import-export-jwk-aes-key.js`) isn't, if the operations are the same? 🤔

olegbespalov: It's the other way around, I think it's just copy-paste issue, I will adjust the `import-export-jwk-aes-key.js` to also have a `try...catch`, which won't harm :+1:

webcrypto/aes.go
```diff
@@ -172,21 +188,31 @@ func (aip *AESImportParams) ImportKey(
...
+	// check the key length
+	var (
+		has128Bits = len(keyData) == 16
+		has192Bits = len(keyData) == 24
+		has256Bits = len(keyData) == 32
+	)
+
+	if !has128Bits && !has192Bits && !has256Bits {
+		return nil, NewError(DataError, fmt.Sprintf("invalid key length %v bytes", len(keyData)))
+	}
```
joanlopez, line 214: Nit; this can be hide behind a method, similar to `isHashAlgorithm`

webcrypto/jwk.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	// we do marshal and unmarshal to get the map of JWK key parameters
+	// where all standard parameters are present, a proper marshaling is done
+	m, err := json.Marshal(sk)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal JWK key: %w", err)
+	}
+
+	// wrap result result into the object that is expected to be returned
```
joanlopez, line 54:
```suggestion
	// wrap result into the object that is expected to be returned
```

webcrypto/subtle_crypto.go
```diff
@@ -633,14 +632,29 @@ func (sc *SubtleCrypto) ImportKey(
...
+		copy(keyBytes, ab)
+	case JwkKeyFormat:
+		var err error
+		keyBytes, err = json.Marshal(keyData.Export())
+		if err != nil {
+			reject(NewError(ImplementationError, "wrong keyData format for JWK format: "+err.Error()))
+			return promise
+		}
+	default:
+		reject(NewError(ImplementationError, "unsupported format "+format))
```
joanlopez, line 656:
```suggestion
		reject(NewError(ImplementationError, "unsupported format "+format))
		return promise
```

Shouldn't we return the promise here as well, as we do for the rest of the error cases? 🤔

webcrypto/subtle_crypto.go
```diff
@@ -759,7 +773,17 @@ func (sc *SubtleCrypto) ExportKey(format KeyFormat, key goja.Value) *goja.Promis
...
 
-		resolve(rt.NewArrayBuffer(result))
+		if format != RawKeyFormat {
+			resolve(result)
+			return
+		}
+
+		b, ok := result.([]byte)
+		if !ok {
+			reject(NewError(ImplementationError, "for "+format+" []byte expected as result"))
```
joanlopez, line 783: Similar to the comment above, should we return here right after the `reject`? 🤔

joanlopez (commented):
It looks generally good, great work! 🎉  

I left a few minor comments, but almost none of them is blocking, except the ones on `webcrypto/subtle_crypto.go`, which I'm wondering if needed for correctness (correct behavior).

Thanks!

webcrypto/subtle_crypto.go
olegbespalov: Yes, we should! Great catch!

joanlopez (approved):

Issue #49, opened 2023-04-24
Implement support for ECDH across the supported APIs


oleiade:
The scope of this issue is to implement support for the ECDH algorithm across the currently supported APIs.

```[tasklist]
## Checklist
- [ ] Implement support for ECDH in `crypto.subtle.generateKey`
- [ ] Implement support for ECDH in `crypto.subtle.importKey`
- [ ] Implement support for ECDH in `crypto.subtle.exportKey`
- [ ] Implement support for  ECDH in `crypto.subtle.sign`
- [ ] Implement support for ECDH in `crypto.subtle.verify`
```

olegbespalov:
Closing the issue. Support of the Derive Keys will be tracked in #10

PR #67, opened 2024-04-09
ECDH support

olegbespalov:
# What?

This PR brings base support for the ECDH into the xk6-webcrypto.

It contains:
* Key pair generation
* Importing/exporting public/private keys (raw and PKCS8 formats) 
* derive bits operation
* WebPlatform API test case for the importing/exporting keys

Part of #49 

Note: the WebPlatform API test case for the derive bits is out of the scope of the PR (but will be done straight after the merging of this PR) to reduce the scope of the PR. It's also good to mention that following #69 is dependent on this change and continued implementation at some aspects (e.g. `spki` support)

The derived key operation is also out of the scope of reducing the PR size.

# Why?

ECDH is part of the WebCrypto API


examples/derive_bits/derive-bits-ecdh.js
```diff
@@ -0,0 +1,58 @@
+import { crypto } from "k6/x/webcrypto";
+
+export default async function () {
+  try {
```
mstoykov, line 4: Do we really need the `try` in these examples? 

I think it will be simpler if it is missing it and it will be just as good of an example

webcrypto/subtle_crypto_test.go
```diff
@@ -70,13 +70,13 @@ func TestSubtleCryptoGenerateKey(t *testing.T) {
 func TestSubtleCryptoImportExportKey(t *testing.T) {
 	t.Parallel()
 
-	t.Run("symmetric", func(t *testing.T) {
+	t.Run("importKeys", func(t *testing.T) {
```
mstoykov, line 73: Is there a reason for this sub test ?

Seems like either the arguments to `executeTestScripts` should be in two subtest or this sutest should just go.

webcrypto/subtle_crypto.go
```diff
@@ -597,11 +605,69 @@ func (sc *SubtleCrypto) DeriveKey(
...
+			}
+
+			return b[:length/8], nil
+		}()
+		if err != nil {
+			reject(err)
+			return
+		}
+
+		resolve(rt.NewArrayBuffer(result))
```
mstoykov, line 667: `rt.NewArrayBuffer` uses the runtime , but this is done in the event loop.

Given your code I would recommend moving away from `promises.New` to `goja.NewPromise` and calling `vu.RegisterCallback` just before the goroutine creation. Then calling from the if below in hte callback as in
```suggestion
callback(func() error {
		if err != nil {
			reject(err)
			return
		}

		resolve(rt.NewArrayBuffer(result))
})
```

This way this whole thing will be executed on the event loop and you can both use the runtime but also the not wrapped resolve/reject.

mstoykov (changes requested):
LGTM in general, I have left some questions and there is one place where there is usage of the runtiem *off* the event loop that is the requested change

examples/derive_bits/derive-bits-ecdh.js
olegbespalov: I've aligned it with the existing examples. However, I believe you're right, and it's better to migrate the whole set of examples (I also wanted to format them all) to `then` in a single PR, which I can do separately. How does it sound?

webcrypto/subtle_crypto_test.go
olegbespalov: yep, you're right will split them

webcrypto/tests/generateKey/failures.js
```diff
@@ -36,15 +36,15 @@ function run_test(algorithmNames) {
         {name: "AES-CBC",  resultType: "CryptoKey", usages: ["encrypt", "decrypt", "wrapKey", "unwrapKey"], mandatoryUsages: []},
         {name: "AES-GCM",  resultType: "CryptoKey", usages: ["encrypt", "decrypt", "wrapKey", "unwrapKey"], mandatoryUsages: []},
         {name: "AES-KW",   resultType: "CryptoKey", usages: ["wrapKey", "unwrapKey"], mandatoryUsages: []},
-        {name: "HMAC",     resultType: CryptoKey, usages: ["sign", "verify"], mandatoryUsages: []},
+        {name: "HMAC",     resultType: "CryptoKey", usages: ["sign", "verify"], mandatoryUsages: []},
```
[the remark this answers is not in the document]
mstoykov: I am just trying to figure out if it is relevant? As in - if this is not changed, will this still work?

What is the original one in the original test file?

webcrypto/subtle_crypto.go
olegbespalov: Let me clarify :thinking: 

 Is the idea to go out of the vanilla go routine because we still use runtime? (`rt.NewArrayBuffer`)

olegbespalov: The clarification question still present, but the change is here https://github.com/grafana/xk6-webcrypto/pull/67/commits/a15ebeb484aa65e6c807461b8161b5fabca1aadd

webcrypto/tests/generateKey/failures.js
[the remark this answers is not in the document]
olegbespalov: I've reverted the change

oleiade (approved):
Looking good, great job 👏🏻  I'm really happy to see ECDH land in the extension  🚀

webcrypto/subtle_crypto.go
mstoykov: the problem is that `rt.NewArrayBuffer` is *not* happening on the event loop in the original code. It happens on a goroutine that isn't running on the event loop. At which point you have a race as this is accessing the runtime, again off the event loop. 

`promises.New` is a wrapper around `goja.NewPromises` and the event loop, so that the `resolve` and `reject` calls can be called on the event loop without the users specifically working with `RegisterCallback` and co.

In this case though as the argument to resolve needs to be an array buffer - you already need to run the `rt.NewArryaBuffer` part *on* the event loop - so you need to use `RegisterCallback` either way. 

Also in this case as *before* you make the goroutine you are on the event loop, the usefulness of `promises.New`'s `reject` being able to be called off the event loop isn't useful and neither is the calling of `resolve` if you need to *again* go back to the event loop to make the ArrayBuffer.

As such my proposal was to *not* use it at all and instead use RegisterCallback.

```pseudo-go
func something() promise {
  // here we are on the event loop
  p, res, rej := promises.New()
  if err != nil {
    rej(err); // this now actually just queues a job on the eventloop, and then when the callstack is empty will actuall reject the promise. 
   // not a deal breaker by any problem, but 
    return p;
  }

  p, res, rej := goja.NewPromise();

  if err != nil {
    rej(err); // this actually just rejects it outright, so the event loop doesn't need to go through this
    return p 
  }
  callback := vu.RegisterCallback()
  go func() {
    // do async work no actual usage of `rt`
   callback(func() error {
     if err !=nil {
       rej(err); // this already is on the event loop as it is the callback - so no need to queue a new job
       return
    }
    res(rt.newArrayBuffer()); // here we are on the event loop so we can use rt no problem.
  }) 
  }
  return p
}
```

In the above case if you drop `promises.New` - you will also never need to do `RegisterCallback` and directly interact with the event loop in the case you have an error.

tl;dr: You can still use `promises.New` - you just will make additional eventloop jobs that will not be useful and you in practice still need to `RegisterCallback` . If we didn't need to do work with the `runtime` before we can resolve/reject the promise - we could've used `promises.New`

olegbespalov: Okay, if I'm reading this right, the https://github.com/grafana/xk6-webcrypto/commit/a15ebeb484aa65e6c807461b8161b5fabca1aadd changes are fine, but we also need to refactor a bit other places where `rt` accesses inside the plain go routines, e.g.:

* https://github.com/grafana/xk6-webcrypto/blob/main/webcrypto/subtle_crypto.go#L193
* https://github.com/grafana/xk6-webcrypto/blob/main/webcrypto/subtle_crypto.go#L293
* etc

mstoykov: Looking at the changes - you now do not have a gorotuine - you directly use the callback, so nohting happens of the event loop .

Otherwise - yes the those are other cases of the same problem that we haven't caught.

olegbespalov: How about this https://github.com/grafana/xk6-webcrypto/pull/67/commits/411de407ede2efd2e51547500d8d9c91cc291dcc ? reject is still out of the callback, but my understanding this should be fine :thinking:

olegbespalov: Okay, as a summary of the internal conversation with @mstoykov 

We should access the runtime (`rt`) only through the event loop (in this example, it's inside the `callback`), but if we directly use `rt.NewPromise` the both `resolve` and `reject` should be executed on top of the event loop, so the result version https://github.com/grafana/xk6-webcrypto/pull/67/files#diff-9e8a31e9776cf152256880a32e1c5751bf67322c23c0cb096d71cf0633d8fe14R665-R673

That way, we'll be protected from the possible races.

mstoykov (approved):

examples/derive_bits/derive-bits-ecdh.js
olegbespalov: Ref. https://github.com/grafana/xk6-webcrypto/pull/77

Issue #41, opened 2023-04-03
Reactivate `sign`/`verify` HMAC tests relying on ECDSA key generation


oleiade:
Some of the tests of the `sign` and `verify` operations using the HMAC algorithm rely on `generateKey` supporting ECDSA. As of today those tests are commented out as the ECDSA algorithm is not yet supported. Once it is, we should uncomment those tests.

PR #68, opened 2024-04-10
HMAC fix for the sign and verify

olegbespalov:
# What?

This change fixes a degradation that happened after merging the https://github.com/grafana/xk6-webcrypto/pull/61

I found it while working on the ECDSA support, and it seems like ECDSA was also a reason why the test case was disabled https://github.com/grafana/xk6-webcrypto/issues/41. 

# Why?

Without the change our example erroring with:
```
./k6 run examples/sign_verify/sign-verify-hmac.js

INFO[0000] {"name":"InvalidAccessError","message":"key algorithm does not describe a HMAC key"}  source=console
```

### DIFF
--- a/webcrypto/subtle_crypto.go
+++ b/webcrypto/subtle_crypto.go
@@ -266,7 +266,7 @@ func (sc *SubtleCrypto) Sign(algorithm, key, data goja.Value) *goja.Promise {
 		// 10.
 		switch normalized.Name {
 		case HMAC:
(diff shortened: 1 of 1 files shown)


mstoykov (approved):
LGTM!

btw usually single method interfaces have `er` at the end of the single method name as their name. That is to say that `hasHash` usually should be `hasher`. 

Not really all that important, but maybe we can fix it in another PR

olegbespalov:
Yep, and I'm trying to follow this go-convention. However, here it's not the `hasher` but something that has a `hash() string` method

Issue #48, opened 2023-04-24
Implement support for ECDSA across the supported APIs


oleiade:
The scope of this issue is to implement support for the ECDSA algorithm across the currently supported APIs.

```[tasklist]
## Checklist
- [ ] Implement support for ECDSA in `crypto.subtle.generateKey`
- [ ] Implement support for ECDSA in `crypto.subtle.importKey`
- [ ] Implement support for ECDSA in `crypto.subtle.exportKey`
- [ ] Implement support for  ECDSA in `crypto.subtle.sign`
- [ ] Implement support for ECDSA in `crypto.subtle.verify`
```

PR #69, opened 2024-04-10
ECDSA support

olegbespalov:
# What?

This PR is based on #67 and extends support of elliptic curves in xk6-webcrypto.

It brings support to the ECDSA for the following methods:

* import/export keys
* sign
* verify

It also brings the `spki` format support for both ECDH and ECDSA.

# Why?

Closes: #48 
Closes: #70


examples/sign_verify/sign-verify-ecdsa.js
```diff
@@ -0,0 +1,49 @@
...
+  for (let i = 0, strLen = str.length; i < strLen; i++) {
+    bufView[i] = str.charCodeAt(i);
+  }
+  return buf;
+};
+
+const printArrayBuffer = (buffer) => {
+  let view = new Uint8Array(buffer);
+  return Array.from(view);
+};
```
mstoykov, line 49: Having seen this in a bunch of places at this point - maybe move them to a helper functio( in another PR)?

webcrypto/subtle_crypto.go
```diff
@@ -263,37 +269,14 @@ func (sc *SubtleCrypto) Sign(algorithm, key, data goja.Value) *goja.Promise {
...
-
-			// 10.
-			mac := hasher.Sum(nil)
-
-			resolve(rt.NewArrayBuffer(mac))
-		default:
-			reject(NewError(NotSupportedError, fmt.Sprintf("unsupported algorithm %q", normalized.Name)))
+		signature, err := signer.Sign(ck, dataToSign)
+		if err != nil {
+			log.Println(err.Error())
```
mstoykov, line 274:
```suggestion
```

leftover debugging log?

webcrypto/subtle_crypto.go
```diff
@@ -1,10 +1,10 @@
 package webcrypto
 
 import (
-	"crypto/hmac"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log"
```
mstoykov, line 7:
```suggestion
```

webcrypto/subtle_crypto.go
```diff
@@ -263,37 +269,14 @@ func (sc *SubtleCrypto) Sign(algorithm, key, data goja.Value) *goja.Promise {
...
-			hasher := hmac.New(hashFn, keyHandle)
-			hasher.Write(dataToSign)
-
-			// 10.
-			mac := hasher.Sum(nil)
-
-			resolve(rt.NewArrayBuffer(mac))
-		default:
-			reject(NewError(NotSupportedError, fmt.Sprintf("unsupported algorithm %q", normalized.Name)))
+		signature, err := signer.Sign(ck, dataToSign)
```
mstoykov, line 272: Shouldn't this be in a goroutine?

webcrypto/subtle_crypto.go
```diff
@@ -372,33 +361,13 @@ func (sc *SubtleCrypto) Verify(algorithm, key, signature, data goja.Value) *goja
...
-			resolve(hmac.Equal(signatureData, hasher.Sum(nil)))
-		default:
-			reject(NewError(NotSupportedError, fmt.Sprintf("unsupported algorithm %q", normalizedAlgorithm.Name)))
+		verified, err := verifier.Verify(ck, signatureData, signedData)
+		if err != nil {
+			reject(err)
+			return
 		}
+
+		resolve(rt.ToValue(verified))
```
mstoykov, line 370:
```suggestion
		resolve(verified)
```

No need for this and also you shouldn't access the runtime off event loop

mstoykov (changes requested):

webcrypto/subtle_crypto.go
olegbespalov: Good catch! we don't use a log, right? Maybe we could add this to the linter check?

examples/sign_verify/sign-verify-ecdsa.js
olegbespalov: I was asking myself the same question but ended up this way since I wanted that example to be "completed" (without the need to have multiple files), just copy/run. Do you think it makes sense?

webcrypto/subtle_crypto.go
olegbespalov: It's in a goroutine

webcrypto/subtle_crypto.go
```diff
@@ -263,37 +268,13 @@ func (sc *SubtleCrypto) Sign(algorithm, key, data goja.Value) *goja.Promise {
...
-			resolve(rt.NewArrayBuffer(mac))
-		default:
-			reject(NewError(NotSupportedError, fmt.Sprintf("unsupported algorithm %q", normalized.Name)))
+		signature, err := signer.Sign(ck, dataToSign)
+		if err != nil {
+			reject(err)
+			return
 		}
+
+		resolve(rt.NewArrayBuffer(signature))
```
mstoykov, line 277: Same problem as in the previous PRs with using the runtime in a goroutine off the event loop

olegbespalov: yeah, but I'd like to do this separately (in ~1h I'll push a PR with fixing all methods)

mstoykov (approved):
LGTM if #75 is merged just after this.

PR #74, opened 2024-04-16
JWK import/export for the ECDH & ECDSA

olegbespalov:
# What?

This PR supports the JWK format for the ECDH and ECDSA.

It also migrates existing JWK usages from the third party (github.com/lestrrat-go/jwx) to SDK usage only.

Note: Initially, I had different commits for the EC JWK and Symmetric JWK, but unfortunately, I squashed them :cry:
Hopefully, that's still fine, keeping in mind the size of the PR.

Based on: https://github.com/grafana/xk6-webcrypto/pull/69

# Why?

* JWK is part of the specification
* Using SDK only is better from the maintenance and security perspective


mstoykov (approved):

oleiade (approved):
Good call on the JWK dependency drop 🙇🏻 🚀

PR #75, opened 2024-04-17
Refactor WebCrypto to be tread safe: part 1

### LINKED ISSUE #71: Migrate from the direct runtime access throw the go routines
# What?

During the https://github.com/grafana/xk6-webcrypto/pull/67#discussion_r1562260215 we figured out that it's essential to not have access to the goja runtime inside the go routines because that could lead to the race.

There are places in the module where we have such access; we should migrate them accordingly:

* https://github.com/grafana/xk6-webcrypto/blob/main/webcrypto/subtle_crypto.go#L193
* https://github.com/grafana/xk6-webcrypto/blob/main/webcrypto/subtle_crypto.go#L293

# Why?

Having access to the throw routine could lead to the race.

olegbespalov:
# What?

This change centralizes and unifies logic and ensures that runtime (`tr`) isn't used outside the event loop.

I also removed references to WebCrypto steps, like (`1.`, `2.`, etc.) since the issue is that for different algorithms, they could mean different steps, which could confuse future maintainers.

Splitting the PR into the parts to make it smaller.

# Why?

Updates #71

### DIFF
--- a/webcrypto/subtle_crypto.go
+++ b/webcrypto/subtle_crypto.go
@@ -219,61 +219,74 @@ func (sc *SubtleCrypto) Decrypt(algorithm, key, data goja.Value) *goja.Promise {
 // The `data` parameter should contain the data to be signed.
 func (sc *SubtleCrypto) Sign(algorithm, key, data goja.Value) *goja.Promise {
 	rt := sc.vu.Runtime()
-	promise, resolve, reject := promises.New(sc.vu)
 
-	// 2.
-	// We obtain a copy of the key data, because we might need to modify it.
-	dataToSign, err := exportArrayBuffer(rt, data)
-	if err != nil {
-		reject(err)
-		return promise
-	}
+	var (
+		dataToSign []byte
+		ck         CryptoKey
+		signer     SignerVerifier
+	)
 
-	// 3.
-	normalized, err := normalizeAlgorithm(rt, algorithm, OperationIdentifierSign)
-	if err != nil {
-		reject(err)
-		return promise
-	}
+	err := func() error {
+		var err error
+		// 2.
+		// We obtain a copy of the key data, because we might need to modify it.
+		dataToSign, err = exportArrayBuffer(rt, data)
+		if err != nil {
+			return err
+		}
 
-	signer, err := newSignerVerifier(rt, normalized, algorithm)
-	if err != nil {
-		reject(err)
-		return promise
-	}
+		// 3.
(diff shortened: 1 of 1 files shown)


webcrypto/subtle_crypto.go
```diff
@@ -36,73 +37,72 @@ type SubtleCrypto struct {
...
-	// 2.
-	// We obtain a copy of the key data, because we might need to modify it.
-	plaintext, err := exportArrayBuffer(rt, data)
-	if err != nil {
-		reject(err)
-		return promise
-	}
+	var plaintext []byte
+	var ck CryptoKey
+	var encrypter EncryptDecrypter
```
mstoykov, line 47: I am suprised gofmt/gofumpt doesn't reformat this to a block of var :thinking:

mstoykov (approved):

webcrypto/subtle_crypto.go
olegbespalov: But it's good to do that. That way, the "shared" variables will be in the same logical block, going to adjust :+1:

olegbespalov: Updated with the latest force push

mstoykov (approved):

webcrypto/subtle_crypto.go
```diff
@@ -128,71 +130,73 @@ func (sc *SubtleCrypto) Encrypt(algorithm, key, data goja.Value) *goja.Promise {
...
+		decrypter  EncryptDecrypter
+	)
 
-	// 3.
-	normalized, err := normalizeAlgorithm(rt, algorithm, OperationIdentifierDecrypt)
-	if err != nil {
-		reject(err)
-		return promise
-	}
+	err := func() error {
```
oleiade, line 144: I'm not familiar with this pattern of wrapping erroring code in dedicated anonymous' function scope. Just informally, I'd be keen to hear more about its benefits 🙇🏻

oleiade (approved):

webcrypto/subtle_crypto.go
olegbespalov: Sure :+1: 

The key point of this change is to centralize the logic of management promises. Previously, it was easier to miss that we either had to reject or return promises.  

See examples:
* https://github.com/grafana/xk6-webcrypto/pull/61#discussion_r1519621870
* https://github.com/grafana/xk6-webcrypto/pull/61#discussion_r1519623244

Also, it simplifies the operations because at this logical block, we only do operations + with returning almost vanilla errors, but it doesn't do any promise-related operations.

PR #76, opened 2024-04-17
 Refactor WebCrypto to be tread safe: part 2

olegbespalov:
# What?

This is a continuation of the #75

This change centralizes and unifies logic and ensures that runtime (tr) isn't used outside the event loop.

I also removed references to WebCrypto steps, like (1., 2., etc.) since the issue is that for different algorithms, they could mean different steps, which could confuse future maintainers.

### Note

The last peace, `ExportKey,` is left untouched for now, but I recognized that it deserves a dedicated PR since we need to adjust the logic a bit. 

E.g. implement more checks for validation operations like:
```go
algValue := key.ToObject(rt).Get("algorithm") // key.ToObject(rt) could be nill
```
and exporter could be an abstraction like in any other operation.

I'll do this tomorrow/Friday

# Why?

Part of #71

### DIFF
--- a/webcrypto/subtle_crypto.go
+++ b/webcrypto/subtle_crypto.go
@@ -725,7 +725,7 @@ func (sc *SubtleCrypto) DeriveBits(algorithm goja.Value, baseKey goja.Value, len
 //     `ALGORITHM` is the name of the algorithm.
 //   - for PBKDF2: pass the string "PBKDF2"
 //   - for HKDF: pass the string "HKDF"
-func (sc *SubtleCrypto) ImportKey(
+func (sc *SubtleCrypto) ImportKey( //nolint:funlen // we have a lot of error handling
 	format KeyFormat,
 	keyData goja.Value,
 	algorithm goja.Value,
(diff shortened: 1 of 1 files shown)


mstoykov (approved):

webcrypto/subtle_crypto.go
```diff
@@ -496,46 +496,66 @@ func (sc *SubtleCrypto) Digest(algorithm goja.Value, data goja.Value) *goja.Prom
...
-	keyGenerator, err := newKeyGenerator(sc.vu.Runtime(), normalized, algorithm)
+		keyGenerator, err = newKeyGenerator(rt, normalized, algorithm)
+		if err != nil {
+			return err
+		}
+
+		return nil
+	}()
+
+	promise, resolve, reject := rt.NewPromise()
```
joanlopez, line 517: If I'm not wrong, with the pattern you're using here, I think you can just use `promises.New` instead, which might make it easier to write a slightly shorter and clearer code (as you don't need to explicitly call `callback` but just `resolve` or `reject`).

Isn't that right? @mstoykov 🤔

joanlopez (approved):

PR #77, opened 2024-04-18
Unify, format, and run tests against examples

olegbespalov:
# What?

* Format examples according the k6-docs standard (https://github.com/grafana/k6-docs/blob/main/.eslintrc)
* It also runs the test against these examples

Note that: import/export examples are commented on since they are actually catch a data race that should be resolved after #76 

Sorry for the big PR. It's better to check by commit; the first commit could be skipped since it's auto-formating.

# Why?

Unifying examples should improve the documentation and simplify their maintenance.

A pull-request starter https://github.com/grafana/xk6-webcrypto/pull/67#discussion_r1561013838

Running the tests against examples should help with finding degradations like https://github.com/grafana/xk6-webcrypto/pull/68


examples/derive_bits/derive-bits-ecdh.js
```diff
@@ -1,55 +1,93 @@
 import { crypto } from "k6/x/webcrypto";
 
 export default async function () {
-  try {
-    // Generate a key pair for Alice
-    const aliceKeyPair = await crypto.subtle.generateKey(
+  // Generate a key pair for Alice
+  crypto.subtle
+    .generateKey(
```
mstoykov, line 6: Is there particular reason for removing `await` + `try/catch` through out the examples?

examples/import_export/import-ecdh-key.js
```diff
@@ -18,46 +18,74 @@ export default async function () {
...
-  );
+  ).then(
+    (aliceKeyPair) => {
+      importKeys(
+        new Uint8Array([
+          4, 218, 134, 37, 137, 90, 68, 101, 112, 234, 68, 87, 110, 182, 85,
+          178, 161, 106, 223, 50, 150, 9, 155, 68, 191, 51, 138, 185, 186, 226,
+          211, 25, 203, 96, 193, 213, 68, 7, 181, 238, 52, 154, 113, 56, 76, 86,
+          44, 245, 128, 194, 103, 14, 81, 229, 124, 189, 13, 252, 138, 98, 196,
+          218, 39, 34, 42,
```
mstoykov, line 29: I feel like this will be better off going in a variable maybe  at the bottom of the script  so that the examples do not have this huge arrays. 

Chances are that people will read them from somewhere either way.

examples/derive_bits/derive-bits-ecdh.js
olegbespalov: Well, it started with https://github.com/grafana/xk6-webcrypto/pull/67#discussion_r1561013838 :sweat_smile: 

It seems like having the `then` chains better catch errors for the precise operations, doesn't it?

mstoykov: Aaah ... sorry, I should've been more explicit :sweat: 

In that comment I meant that we don't do anything with the exception ... so catching it just to log it seems bad instead of letting it propagate. Also makes the example more indented without any other benefit.

But movign to `then` from `await` seems to go in the exact opposite direction.

I guess we can really drop the `try`/`catch` if they are only used for this, but moving away from `await` doesn't seem to improve readability in my case

olegbespalov: Personally, the `then` syntax seems better to me, but I don't really have a strong opinion here :shrug: 

I see the benefits of unifying since it's useful for the https://github.com/grafana/xk6-webcrypto/pull/77/commits/806d97e8652554760d2f0a3cd17338c41e95b000 where we check for the key words `got unexpected error,` but on the other side, we could stick with the something standard.

olegbespalov: Reverted with the latest force push

mstoykov (approved):

olegbespalov:
@mstoykov FYI, after including import/export, I've found that one of the examples become broken (:tada:) and I include a minor fix [445ad03](https://github.com/grafana/xk6-webcrypto/pull/77/commits/445ad03a5689ee0641f8507d444f8165fc223803) to this PR. Hope that works!

olegbespalov:
@joanlopez Since this PR is mostly about auto-applying the eslint + moving things around, I'm going to merge it as it is (since I need a couple more based on it). If you have any other feedback, feel free to bring it, and I'll iterate on it separately with PRs if needed.

Issue #84, opened 2024-10-14
Issue with `normalizeAlgorithm()` universally uppercasing algorithm names


Billyzou0741326:
The comment here isn't true for `RSASSA-PKCS1-v1_5` (note the lower case `v` in the canonical algorithm name):

https://github.com/grafana/xk6-webcrypto/blob/f68083cf1c7c8565e8d742f6f43034eac788dcd4/webcrypto/algorithm.go#L148-L151

`RSASSA-PKCS1-v1_5` isn't currently supported, so there's not necessarily a problem with the current implementation. 

But nonetheless, seeing how there are usages of `normalizeAlgorithm()` in validating against user input (in `SubtleCrypto.Encrypt` and `SubtleCrypto.Decrypt`) as well as selecting the right `SignerVerifier`, `KeyGenerator`, `KeyImporter`, etc (as I'm typing this out, I realized it's used pretty much everywhere), it's definitely an error to uppercase the algorithm name in the case of `RSASSA-PKCS1-v1_5` in those use cases.

To keep the current behavior backward compatible while paving the way for future `RSA` implementation, perhaps it's good to make a special case for `RSASSA-PKCS1-v1_5` in `normalizeAlgorithm()`?  Another way is to test for `strings.ToUpper(RSASsaPkcs1v15)`, but that just feels wrong.

mstoykov:
Hi @Billyzou0741326 , thank you for reporting this 🙇 

Checking this it seems you are correct and that the correct way to test for this is actually to have a list of supported algorithms and their canonical names that we match against *case insensitively*. 

Your proposed solution also will work for this case, but I wonder if this doesn't happen in other places as well. Maybe just updating it in normalizeAlgorithm is enough 🤷 .


Given the fact this algorithm isn't currently supported I don't know whether we should fix it now as part of its implementation. But I guess we can at least apply this smaller patch

cc @olegbespalov @oleiade Are there any plans on working in RSA<anything> soon ™ ?

Billyzou0741326:
I could be wrong, but I think as far as algorithm name matching, it's mostly done up front in `normalizedAlgorithm()`, and the result of that is passed down to be used elsewhere. So, having the patch applied in `normalizedAlgorithm()` should be sufficient.

Having done some preliminary testing on rsa implementation (mostly for my immediate needs), the small snippet of patch that I mentioned above should solve the issue for `RSASSA-PKCS1-v1_5` which again, is only relevant if the implementation is there.

[...]

Issue #47, opened 2023-04-24
Implement support for RSA across the supported APIs


oleiade:
The scope of this issue is to implement support for the RSA algorithms across the currently supported APIs.

```[tasklist]
## Checklist

- [ ] Implement support for RSA in `crypto.subtle.generateKey`
- [ ] Implement support for RSA in `crypto.subtle.importKey`
- [ ] Implement support for RSA in `crypto.subtle.exportKey`
- [ ] Implement support for RSA in `crypto.subtle.encrypt`
- [ ] Implement support for RSA in `crypto.subtle.decrypt`
- [ ] Implement support for RSA in `crypto.subtle.sign`
- [ ] Implement support for RSA in `crypto.subtle.verify`
```

PR #85, opened 2024-10-22
Support of RSA 

olegbespalov:
# What?

This PR brings support for the RSA algorithms across available APIs.

Sorry for the huge PR, but the predominant issue is that to make RSA useful (sign, verify, encrypt, decrypt), we need to have a lot of functionality already (generation, export, import), so it's kind of all related. However, I did break pieces in commits, so you could try checking them.

The vital highlight could be that for the RSA-PSS algorithm, we don't support saltLength 0 (since golang SDK doesn't support it). So, I'm thinking of also writing a log message whenever a user uses this to bring attention. But this will be in a separate PR.

# Why?

Closes: #47 
Closes: #84


Billyzou0741326:
I think this closes #84 too. Thanks for the quick implementation

webcrypto/algorithm.go
```diff
@@ -145,16 +146,30 @@ func normalizeAlgorithm(rt *sobek.Runtime, v sobek.Value, op AlgorithmIdentifier
...
 	// Algorithm identifiers are always upper cased.
 	// A registered algorithm provided in lower case format, should
 	// be considered valid.
-	algorithm.Name = strings.ToUpper(algorithm.Name)
+	name = strings.ToUpper(name)
 
-	if !isRegisteredAlgorithm(algorithm.Name, op) {
-		return Algorithm{}, NewError(NotSupportedError, "unsupported algorithm: "+algorithm.Name)
+	// exception is made for RSASSA-PKCS1-v1_5
+	if name == strings.ToUpper(RSASsaPkcs1v15) {
```
oleiade, line 168: 👍

webcrypto/subtle_crypto.go
```diff
@@ -446,7 +446,7 @@ func (sc *SubtleCrypto) Digest(algorithm sobek.Value, data sobek.Value) *sobek.P
 		var ok bool
 		hashFn, ok = getHashFn(normalized.Name)
 		if !ok {
-			return NewError(NotSupportedError, "unsupported algorithm: "+normalized.Name)
+			return NewError(NotSupportedError, "unsupported algorithm, can't map hash function: "+normalized.Name)
```
oleiade, line 449: Nit, as we're at it we could go full explicit and help the user understand how to solve the issue, what do you think of something along the lines of the following?

```suggestion
			return NewError(
				NotSupportedError,
				"unsupported digest algorithm '"+normalized.Name+"', "+
				"accepted values are: SHA-1, SHA-256, SHA-384, and SHA-512",
			)
```

webcrypto/key.go
```diff
@@ -188,8 +188,10 @@ func newKeyGenerator(rt *sobek.Runtime, normalized Algorithm, params sobek.Value
 		kg, err = newHMACKeyGenParams(rt, normalized, params)
 	case ECDH, ECDSA:
 		kg, err = newECKeyGenParams(rt, normalized, params)
+	case RSASsaPkcs1v15, RSAPss, RSAOaep:
+		kg, err = newRsaHashedKeyGenParams(rt, normalized, params)
 	default:
-		return nil, errors.New("key generation not implemented for algorithm " + normalized.Name)
+		return nil, NewError(NotImplemented, "unsupported algorithm for key generation: "+normalized.Name)
```
oleiade, line 194: Nit, same approach as for the digest operation, we could try to make the error message even more explicit.

```suggestion
				validAlgorithms := []string{AESCbc, AESCtr, AESGcm, AESKw, HMAC, ECDH, ECDSA, RSASsaPkcs1v15, RSAPss, RSAOaep}
		return nil, NewError(
			NotImplemented,
			"unsupported key generation algorithm '"+normalized.Name+"', "+
				"accepted values are: "+strings.Join(validAlgorithms, ", "),
		)
```

oleiade (approved):
Awesome work 👏🏻 🚀 Feels great to see RSA coming to k6. Left a couple of absolutely disposable nits, but looks all good to me 🙇🏻

webcrypto/tests/wpt-patches/WebCryptoAPI__sign_verify__rsa.js.patch
```diff
@@ -0,0 +1,39 @@
...
+index 5abadd3d4..c1673065a 100644
+--- a/WebCryptoAPI/sign_verify/rsa.js
++++ b/WebCryptoAPI/sign_verify/rsa.js
+@@ -1,8 +1,8 @@
+ 
+ function run_test() {
+-    setup({explicit_done: true});
++    // setup({explicit_done: true});
+ 
+-    var subtle = self.crypto.subtle; // Change to test prefixed implementations
```
mstoykov, line 11: I wonder if it won't be easier to actually set `self` to `globalThis` which seems to be what it meant to do https://developer.mozilla.org/en-US/docs/Web/API/Window/self

webcrypto/subtle_crypto_test.go
```diff
@@ -167,6 +183,23 @@ func TestSubtleCryptoEncryptDecrypt(t *testing.T) {
...
+	t.Run("RSA-OAEP", func(t *testing.T) {
+		t.Parallel()
+
+		ts := newWebPlatformTestRuntime(t)
+
+		gotErr := ts.EventLoop.Start(func() error {
+			err := executeTestScripts(ts.VU.Runtime(), webPlatformTestSuite+"encrypt_decrypt", "rsa_vectors.js", "rsa.js")
+			require.NoError(t, err)
+
+			_, err = ts.VU.Runtime().RunString(`run_test()`)
```
mstoykov, line 196: nit: a lot of this test semm ... the same apart from the name and the some strings here.

It might be .. .better to make them into a table test?

webcrypto/key.go
olegbespalov: I agree that it's nice to make error messages actionable, but frankly, I'm not sure if the local list serves here the best. I'll try to iterate on that in the following PRs by combining the details from `isRegisteredAlgorithm`, which could serve as the single source of truth

webcrypto/subtle_crypto_test.go
olegbespalov: Sure, let me address this after merging this and https://github.com/grafana/xk6-webcrypto/pull/87 as part as clean up

Issue #81, opened 2024-05-02
Downloading Web Platform Tests


olegbespalov:
# What?

Currently, we store Web Platform Tests under the git in the repository:
https://github.com/grafana/xk6-webcrypto/tree/main/webcrypto/tests

Wee could instead download them, apply the patch if needed, and then run the tests against them.

Such approach was chosen during the https://github.com/grafana/k6/pull/3696

# Why?

It should improve the maintenance of the Web Platform Tests and potentially simplify the process of incorporating updates.

PR #87, opened 2024-10-25
Web Platform Tests Checkout

olegbespalov:
# What?

This PR implements running of the web platform tests not against static in-repository files, but it does check out to the latest proven to commit, apply patches with k6/web crypto implementation specific, and runs test against it.

It also brings tooling around, like easy checkout and patch generation.

# Why?

This approach has proven successful in https://github.com/grafana/k6/pull/3696 and has many benefits. For example, we can clearly identify the difference between K6's and WPT test suites.

It also improves developer (and reviewer) UX.

Closes: #81


oleiade:
I'll name @joanlopez for this review in my place if that's okay with you folks, as he drove this work on streams before and would likely have much more insights to offer than me 🙇🏻

webcrypto/subtle_crypto_test.go
```diff
@@ -35,10 +46,10 @@ func TestSubtleCryptoGenerateKey(t *testing.T) {
 	t.Run("successes", func(t *testing.T) {
 		t.Parallel()
 
-		ts := newConfiguredRuntime(t)
+		ts := newWebPlatformTestRuntime(t)
 
 		gotErr := ts.EventLoop.Start(func() error {
-			err := executeTestScripts(ts.VU.Runtime(), "./tests/generateKey", "successes.js")
+			err := executeTestScripts(ts.VU.Runtime(), webPlatformTestSuite+"generateKey", "successes.js")
```
mstoykov, line 52: I pointed out this in the RSA PR, but a lot of those test seem *very* similar and likely can be a table test instead.

webcrypto/tests/wpt-patches/WebCryptoAPI__derive_bits_keys__ecdh_bits.js.patch
```diff
@@ -0,0 +1,39 @@
+diff --git a/WebCryptoAPI/derive_bits_keys/ecdh_bits.js b/WebCryptoAPI/derive_bits_keys/ecdh_bits.js
+index 36b29c20a..e3bd4ccb5 100644
+--- a/WebCryptoAPI/derive_bits_keys/ecdh_bits.js
++++ b/WebCryptoAPI/derive_bits_keys/ecdh_bits.js
+@@ -1,7 +1,7 @@
+ 
+ function define_tests() {
+     // May want to test prefixed implementations.
+-    var subtle = self.crypto.subtle;
++    var subtle = crypto.subtle;
```
mstoykov, line 10: Again I comment on this in the RSA test , but we can set `globalThis.self` to `globalThis` and won't need to do this everywhere.

webcrypto/tests/wpt-patches/WebCryptoAPI__digest__digest.https.any.js.patch
```diff
@@ -0,0 +1,14 @@
...
+index 379d9311f..949b1669f 100644
+--- a/WebCryptoAPI/digest/digest.https.any.js
++++ b/WebCryptoAPI/digest/digest.https.any.js
+@@ -119,9 +119,6 @@
+     });
+ 
+ 
+-    done();
+-
+-
```
mstoykov, line 11: This seems very strange ... Can we maybe have comments in the places where we do strange changes :)

mstoykov (approved):
Thank you for doing this nad LGTM in general.

I would really prefer if every change has a comment on *why* we are doing this. Some do whcih is great, but some that seem "strange" do not and I wonder to what extend that is "this is not needed for us" and "we have not implemented something" or "we implemented something wrongly". 

I would argue (as I have done for other test suties including tc39's test262) to just add empty implementations in cases where we do not have to do something special for example, instead of changingthe tests

olegbespalov:
>I would really prefer if every change has a comment on why we are doing this. Some do whcih is great, but some that seem "strange" do not and I wonder to what extend that is "this is not needed for us" and "we have not implemented something" or "we implemented something wrongly".

I do accept all the comments and will be happy to implement them, but probably as part of the following PR since I'd like to merge firstly this and RSA support. I will open take anyway :+1:

olegbespalov:
@joanlopez I've merged to run the PRs train. Feel free to drop more feedback here since in any case, soon I'll open one more to address Mihails suggestions

joanlopez (commented):
> @joanlopez I've merged to run the PRs train. Feel free to drop more feedback here since in any case, soon I'll open one more to address Mihails suggestions

Hey @olegbespalov! I did reviewed these changes and I don't have any comment that isn't covered yet by Mihail's suggestions, so 👌🏻 Great work! 🚀

PR #88, opened 2024-10-25
Apply fix for the async interrupt

olegbespalov:
# What?

Include the k6 version with async interrupt fix

# Why?

https://github.com/grafana/k6/pull/4017


webcrypto/subtle_crypto.go
```diff
@@ -87,7 +87,7 @@ func (sc *SubtleCrypto) Encrypt( //nolint:dupl // we have two similar methods
 		return nil
 	}()
 
-	promise, resolve, reject := rt.NewPromise()
+	promise, resolve, reject := newPromise(rt)
```
mstoykov, line 90: Given how this is used ... I think it mgith be better to actually fix this more in place.

You do use the callback quite a lot here in htis code so it is probably better if that returns the error from resolve and reject in each place.

The line just below here seems to be the bigger problem, but I would either be from this also going through the event loop and returning there or only panicking in this place.

Panics in general have this bad case where it seems like hte code will continue downwards but it actually doesn't. So I really prefer not to use them too much.

it is unfortunate that Sobek/goja is build with them for exceptions, but that doesn't mean we should have panics everywhere.

Also I am not certain what happens if there is panic on the event loop, but I would expect it is probably not handled great if Sobek/goja doesn't make it into an error.

mstoykov (approved):

oleiade (approved):

PR #90, opened 2024-10-25
Web Platform Tests part 2

olegbespalov:
# What?

This primarily applies suggestions that @mstoykov made in #87.

It also moves tests around, and the reason is to have the chance to test suites separately in CI (before the PR, we run the regular tests twice)

# Why?

This simplifies the maintenance of the tests and improves the developer's UX.


olegbespalov:
Also @mstoykov huge thanks for the suggestion about the approach, and @joanlopez, for the implementation that I used as the base, I really love how now everything is structured :heart_eyes:

webcrypto/tests/subtle_crypto_test.go
```diff
@@ -0,0 +1,197 @@
+// Part of the Web Platform Tests suite for the k6's WebCrypto API
+//go:build wpt
+
+package tests
```
olegbespalov, line 4: Maybe also explicitly name this package `wpt`?

webcrypto/tests/util/overloads.js
```diff
@@ -0,0 +1,23 @@
...
+
+// this is a minimal implementation of the done function
+// which used in many web platform tests
+function done() {}
+
+// this is a minimal implementation of the setup function
+function setup() {}
+
+// some tests use the global object, so we need to define it
+const self = globalThis;
```
mstoykov, line 23: I would do `globalThis.<identifier>` if I would like them to be globally available as that makes it more explicit and it will work if this is called as a module in some future case.

webcrypto/tests/util/testharness.js
```diff
@@ -91,13 +91,3 @@ function assert_in_array(actual, expected, description) {
 function assert_unreached(description) {
     throw `reached unreachable code, reason: ${description}`
 }
```
mstoykov, line 93: Do we need this "adaptation" ? can we have a diff for it as well?

mstoykov (approved):
LGTM! I left two blocking comments that we can stream line it even more.

webcrypto/tests/util/testharness.js
olegbespalov: Good question, it's something from the original implementation which probably should [cover this](https://github.com/web-platform-tests/wpt/blob/master/resources/testharness.js#L2375-L2379), let me check quickly if it can be dropped

olegbespalov: Well, quickly checking, I think we need this :cry: It's the only place where it's defining, keeping in mind that we don't fully rely on the original `testharness.js` and `helpers.js`

webcrypto/tests/util/overloads.js
olegbespalov: fixed

webcrypto/tests/util/testharness.js
mstoykov: Yeah, but the questions is more "can we just really on the original ones?" 

Or can we have diff patch against the original ones (like we have for the tests) and run them patched instead.

olegbespalov: Probably, yes. Not sure, though, if it's just a simple replacement. Perhaps if it was possible, we started to use them at the moment of beginning the webcrypto. How about opening a separate issue? Ideally, we should use the original test framework as much as possible.

webcrypto/tests/subtle_crypto_test.go
joanlopez: I have no strong opinion, but sounds good! 👍🏻

webcrypto/tests/util/testharness.js
olegbespalov: Okay, after re-looking to streams, it seems like it's not a big deal, so I applied the fixed. It removes the need to any overload (even if test time increased a bit :thinking: )

joanlopez (approved):

mstoykov (approved):

